PR #9, opened 2022-08-12
Feature/3 multiple views

### LINKED ISSUE #3: Javascript File uses a Global `view` variable
Currently, the `arcGisJsInterop.js` file uses a global view variable to keep track of the map view. 

https://github.com/dymaptic/dy-blazor-api-core/blob/98de2d7d0fa6c6e00317a0f636c4497156497718/src/dymaptic.Blazor.GIS.API.Core/wwwroot/arcGisJsInterop.js#L4

I believe that this makes it so that we can't have multiple map views, it is also kind of just asking for trouble I think (having a global variable for this). Can we make this so the reference is just passed back and forth for each function that needs it? Basically can we make all of these functions truly functional and not have any global mutation?

TimPurdum:
closes #3 
1. Added an object collection, `arcGisObjectRefs`, which functions as a `Dictionary<Guid, ArcGisObject>`.
2. Made all methods support multiple views.
3. Moved all references to layers, widgets, and graphics into this collection.
4. Made `MapComponent.DisposeAsync` always call to destroy an object in this collection in JavaScript.
5. Cleaned up a lot of other tracking logic that wasn't necessary anymore.
6. Tested in the public and private samples.
7. Incremented the version numbers to 1.0.1.


src/dymaptic.Blazor.GIS.API.Core/wwwroot/arcGisJsInterop.js
```diff
@@ -203,45 +195,47 @@ export function buildMapView(dotNetReference, long, lat, rotation, mapObject, zo
...
-            if (newWidgets.find(nw => nw.type === aw.type) === undefined) {
-                oldWidgets.push(aw);
-            }
-        });
-        oldWidgets.forEach(ow => {
-            view.ui.remove(ow);
-            activeWidgets.splice(activeWidgets.indexOf(ow), 1);
-            ow?.destroy();
-        });
+        Object.values(arcGisObjectRefs).forEach(o => o?.destroy());
```
morehavoc, line 210: This takes in a viewId, but doesn't seem to filter `arcGisObjectRefs` with that Id, instead, it destroys everything, is that correct?

src/dymaptic.Blazor.GIS.API.Core/wwwroot/arcGisJsInterop.js
```diff
@@ -203,45 +195,47 @@ export function buildMapView(dotNetReference, long, lat, rotation, mapObject, zo
...
+        switch (component?.declaredClass) {
+            case 'esri.Graphic':
+                component.layer.graphics.remove(component);
+                break;
         }
-        unsetWaitCursor();
+        component?.destroy();
+        delete arcGisObjectRefs[componentId];
+        let view = arcGisObjectRefs[viewId];
+        view?.ui?.remove(component);
```
morehavoc, line 229: Is this safe to call even if it is an `esri.Graphic`? (i.e. not part of the UI)

morehavoc (approved):
This is looking great! I'm glad we were able to make these changes without any changes to the exposed API, so no user changes to make all this happy.

PR #18, opened 2022-08-31
Bug/15 firefox

### LINKED ISSUE #15: In Firefox, an error occurs when loading the samples
When browsing to: https://dy-blazor-samples-server.azurewebsites.net/, the initial load works, but then switching to any page on the left menu fails with an error and the page doesn't load.

The initial load works:
image

Clicking on an item in the menu causes an error:
image

TimPurdum:
Closes #15 
Also fixes scrolling into header (previously fixed in private repo for sample projects.


TimPurdum:
Trivial code change, merging

PR #23, opened 2022-09-13
Feature/14 error on missing api key

### LINKED ISSUE #14: Surface an error or log message when the API key isn't found
I was just experimenting with this in WASM and was using a static API key. I had the `appsettings.json` file in the wrong place and so the API key wasn't available. In this case the map just didn't load. It would be nice if there was a log message or an error of some kind so that as the developer I can tell that's what is missing.

TimPurdum:
Closes #14

- Defaults to showing an Error message, which also points to the `UsingTheAPI.md` documentation file.
- Added documentation on how to suppress the error and use the default login screen.
- Added two ways of providing the flag to suppress the error.
- In the documentation, pointed out that the login screen was not as secure as OAuth, and also pointed out the sample code in `MainLayout.razor` for implementing OAuth login.

We will create a separate gissue for a more thorough OAuth implementation.

### DIFF
--- a/src/dymaptic.Blazor.GIS.API.Core/Components/Views/MapView.razor.cs
+++ b/src/dymaptic.Blazor.GIS.API.Core/Components/Views/MapView.razor.cs
@@ -478,8 +478,13 @@ public partial class MapView : MapComponent
             return;
         }
 
-        if (Rendering || (Map is null && WebMap is null) || ViewJsModule is null ||
-            string.IsNullOrWhiteSpace(ApiKey)) return;
+        if (Rendering || (Map is null && WebMap is null) || ViewJsModule is null) return;
+
+        if (string.IsNullOrWhiteSpace(ApiKey))
+        {
+            ErrorMessage = "No ArcGIS API Key Found. See ReadMe.md for instructions on providing an API Key.";
+            StateHasChanged();
+        }
 
         Rendering = true;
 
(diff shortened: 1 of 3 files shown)


src/dymaptic.Blazor.GIS.API.Core/Components/Views/MapView.razor.cs
```diff
@@ -478,8 +491,18 @@ protected override async Task RenderView(bool forceRender = false)
...
         }
 
-        if (Rendering || (Map is null && WebMap is null) || ViewJsModule is null ||
-            string.IsNullOrWhiteSpace(ApiKey)) return;
+        if (Rendering || (Map is null && WebMap is null) || ViewJsModule is null) return;
+
+        if (string.IsNullOrWhiteSpace(ApiKey) && (AllowDefaultEsriLogin is null || !AllowDefaultEsriLogin.Value))
+        {
+            ErrorMessage = "No ArcGIS API Key Found. See UsingTheAPI.md for instructions on providing an API Key or suppressing this message.";
+#if DEBUG
```
morehavoc, line 499: Should we really put the entire `if` block inside the `#if DEBUG`? Since all we are doing is the print, there's no point in even checking if we are not in `DEBUG`.

morehavoc (approved):
See Comment about `#IF DEBUG`

Issue #17, opened 2022-08-31
Widgets must be within a MapView, but that's not a requirement in the JS AP


morehavoc:
In the JS Api, I can put a widget on the map, or I can place it in an arbitrary location within the DOM tree and simply specify the MapView that it should be attached to. For example, I might want to put the basemap switcher under a menu on the left of the map with other menu items for my app (not be limited to placing it only on top of the map view).

In the JS Api, you can specify the DOM node like: 

```javascript
const basemaps = new BasemapGallery({
        view,
        container: "basemaps-container"
      });
```

Today, our API only allows you to position these within the `MapView` (and that's how the reference to the view is created, the widget is a child of the view). This is very powerful, but it would be nice to be able to place these outside the map as well, by providing the link to the view we want it to use.

TimPurdum:
merged with #32

PR #26, opened 2022-09-15
Feature/17 free the widgets

TimPurdum:
This turned out to be a bit harder than I anticipated, and raised an important issue to be aware of.

When a Blazor page loads, it goes through multiple render cycles. On each cycle, the components are recreated. This makes references to _sibling_ components very hard. The `Id` regenerates, and you end up with a stale reference. There are ways around it, but it would all have to be exposed at the consumer level, or require incredible hackery I haven't yet discovered.

So, instead, I opted to require leaving the declared widget _inside_ the `MapView` in razor, which gives it the correct parent-child references, and then give it an external `ContainerId` which points to a div where you want it to actually render. This makes it fairly inline with the use of `OverlayPosition`.

```html
<div id="gallery-box"></div>
<MapView>
    <BasemapGalleryWidget ContainerId="gallery-box" />
</MapView>
```

One other issue I ran into is that _destroying_ a widget also removes the container. I fixed this by wrapping in an extra container in the JavaScript.


TimPurdum:
Closes #17

morehavoc (approved):

Issue #31, opened 2022-09-22
Update Samples Application


TimPurdum:
1. Add a short set of instructions/directions to each sample
2. Center the Scenes.razor to show the actual graphics
3. Alter the basemaps for variety
4. Change the color scheme of graphics to match our design
5. When adding new points to a line or polygon on Drawing.razor, add them at an offset so they show immediately.

PR #35, opened 2022-09-24
Added new basemaps, color theme, borders, instructions.

TimPurdum:
Drawing12

PlaceSelector2
Routing2
ServiceAreas3
SqlFilterQuery2
SqlQuery5
Scene


TimPurdum:
Closes #31

seahro (approved):
Minor color adjustments, new map references, couple async task adjustments...looks good

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ServiceAreas.razor
```diff
@@ -1,8 +1,10 @@
 ﻿@page "/service-areas"
 <PageTitle>Service Areas</PageTitle>
 <h1>Service Areas</h1>
-<a target="_blank" href="https://developers.arcgis.com/javascript/latest/find-service-areas/">ArcGIS API for JavaScript</a>
-
+<a target="_blank" href="https://developers.arcgis.com/javascript/latest/find-service-areas/">ArcGIS API for JavaScript Reference</a>
+<p class="instructions">
+    Click on any point on the map to be shown a set of "Service Areas", drivable distances in 5, 10, and 15 minute bands.
```
codergrl, line 6: Can this be reworded? I'm not sure if Service Areas are drivable distances or if you can see service areas and drivable distances.

codergrl (changes requested):
Made a couple of suggestions for the write-ups. I'd like to also suggest we come up with a different color for links. Right now it's the default link-blue and I'm not sure it works with the rest of the page design.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ServiceAreas.razor
TimPurdum: @codergrl I'm not sure I understand. The definition of Service Area on the linked ArcGIS JS page is
> A service area, also known as an isochrone, is a polygon that represents the area that can be reached when driving or walking on a street network.

codergrl: I see. Ok, how about rewording it to something like this:

Click on any point on the map to be shown a set of "Service Areas". Service Areas are drivable distances in 5, 10, and 15 minute bands.

TimPurdum: sounds good!

TimPurdum:
@codergrl this will require your re-approval to close. I implemented the text changes you suggested, and made links (for now) be the scheme purple, and black on hover. Once we have more design ideas from Chantelle we can finalize these.

codergrl (approved):
👍🏻

Issue #40, opened 2022-09-27
Add LayerList Widget


TimPurdum:
[ArcGIS API for JavaSript - LayerList](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-LayerList.html)
1. Add logic to the `switch` statement in the `arcGisJsInterop.ts` function `addWidget()` to handle the `LayerList` widget type. This should set the following values (if set in C#):
    - `iconClass` (string)
    - `label` (string)
    - `view` - this should be set to the view found at the top of `addWidget()`
2. Add an event handler for the following events that calls back to the dotnet object reference for the `listItemCreatedFunction` event. Use the `Search` widget as an example.
3. Create a new backlog gissue to add `LayerListViewModel` abilities later.
4. Create a file `...Core\Components\Widgets\LayerListWidget.cs` that inherits from `Widget.cs`, and has `public string? IconClass` and `public string? Label` with the `[Parameter]` and `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` attributes.
5. Add the code for the `ListItemCreated` event handler to `LayerList.Widget.cs`, again using `SearchWidget.cs` as a template.
6. Create a new samples page based on [this demo](https://developers.arcgis.com/javascript/latest/sample-code/widgets-layerlist/)

TimPurdum:
Based on discussions with @seahro and @morehavoc , here are some steps for implementation.
1. Create a method `buildDotNetListItem(item: ListItem)` that converts an arcgis `ListItem` to something that the C# json parser can handle.
2. When adding the widget, set `layerListWidget.listItemCreatedFunction` to a new anonymous function, e.g. `layerListWidget.listItemCreatedFunction = (evt) => { };`
3. Implement `ListItem` in C#.
4. Change `OnLayerListSelectResult` and `OnLayerListSelectResultHandler` to `OnListItemSelected` and `OnListItemSelectedHandler` in `LayerListWidget.cs`.
5. Update the call to C# in `listItemCreatedFunction` to something like `widget.DotNetObjectReference.InvokeAsync<DotNetListItem>('OnListItemSelected`, dotNetListItem)` to return the `DotNetListItem` _back_. Verify that the data is getting set in C#.
6. In `LayerListWidget.cs`, change the event handler to `public Func<ListItem, Task<ListItem>>? OnListItemSelectedHandler {get; set;}`.
[...]

PR #48, opened 2022-10-03
incorrect feature number #40 add layerlist widget

seahro:
layer list widget corresponds to gissue #40 .


TimPurdum:
@Chantelle Campbell just FYI, this task should still be open, but we had an issue with branch names that caused the syncing from GitHub to move it.

PR #54, opened 2022-10-07
Release Candidate for GeoBlazor rename

TimPurdum:
I believe that we are in a good state to create a release, and update/rename our nuget package. Versions are set to `1.1.0`. Please give a quick review of anything you think might be urgent before we create a release.


morehavoc (approved):
I've seen all of this in previous work PRs. So many name changes!

seahro (approved):
Name changes look good

PR #57, opened 2022-10-11
Feature/55 docs

TimPurdum:
1. Created a `docs` folder that contains a Jekyll site for publishing via GH Pages.
2. Added a few basic pages like `Getting Started` and `Authentication`.
3. Created a pipeline in the build of `dymaptic.GeoBlazor.Core` that will generate markdown files automatically from XML comments and add them to the `docs\pages\classes`.
4. Added lots of XML comments to public classes and properties.


seahro (approved):
Jekyl autogenerated pages look good

docs/pages/classes/Components
```diff
@@ -0,0 +1,209 @@
+---
```
morehavoc, line 1: @TimPurdum All of these files (in `classes`) are auto-generated, correct? Do we really want to check them all in?

TimPurdum: They are auto-generated, but only by building `dymaptic.GeoBlazor.Core`. So we'd have to customize the GH pages build to generate them. I haven't gotten that far in my knowledge.

morehavoc: I see, okay.

PR #79, opened 2022-10-14
added white on hover and active for visible search results

TimPurdum:
Fixes visibility of dropdown items in docs search when hovering or using cursor to highlight.


seahro (approved):
new button styles

PR #83, opened 2022-10-17
Unique Value sample, implementation, added FeatureLayer properties.

TimPurdum:
1. Added the `UniqueValueRenderer` and related classes.
2. Created a new sample page with US Highway types painted different colors.
3. Added a few properties to `FeatureLayer` required by the sample.

One thing I noticed/haven't figured out. For some reason the docs generation git thinks is modifying all the markdown files, even though my IDE says "Contents are Identical". So you'll see a lot of docs in the PR, but most have no actual changes. I'm curious how this will show when I open the PR.


TimPurdum:
interesting, those extra file commits did not show in GH on the PR.

seahro (approved):
Looks Good.  We will want to add an instruction item to the main readme.  Necessity of having both powershell 7 and the preview program installed on the machine running GeoBlazor.
> https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-windows?view=powershell-7.2#install-powershell-using-winget-recommended

PR #89, opened 2022-10-20
Release Candidate 1.1.1

TimPurdum:


docs/_config.yml
```diff
@@ -0,0 +1,77 @@
...
+# https://learnxinyminutes.com/docs/yaml/
+#
+# Site settings
+# These are used to personalize your new site. If you look in the HTML files,
+# you will see them accessed via {{ site.title }}, {{ site.email }}, and so on.
+# You can create any custom variable you would like, and they will be accessible
+# in the templates via {{ site.myvariable }}.
+
+title: GeoBlazor
+email: info@dymaptic.com
```
morehavoc, line 22: This doesn't need to stop this PR, but we have an email now: `geoblazor@dymaptic.com` and a separate twitter handle!

morehavoc (approved):

docs/_config.yml
TimPurdum: I will add for the next release.

PR #95, opened 2022-10-22
Internet connectivity check and messaging

### LINKED ISSUE #87: Improve error handling around internet disconnect
#78 raised a question about "offline" mode. While we can't support that in the near future (it is a limitation of the ArcGIS JS API), we can improve error handling if the browser detects that it is offline.

1. Expose a clear, human-readable exception when the browser detects that it is offline.
2. Stop spamming recurring messages from an incomplete circuit.
3. Implement actual retry logic. This could be just a few hard-coded retries for now, to be extended to something configurable in the future.

TimPurdum:
Closes #87 
- Check connectivity using `navigator.onLine` and also a call to https://arcgis.com/sharing/rest
- Added a clear message in the map container if offline mode is detected

image


src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -1066,4 +1068,32 @@ function buildDotNetListItem(item: ListItem): DotNetListItem | null {
...
+function checkConnectivity(viewId) {
+    let connectError = new Error('Cannot connect to ArcGIS Services!');
+    let message = '<div><h1>Cannot connect to ArcGIS Services.</h1><h2>Please check your internet connection.</h2></div>';
+    let mapContainer = document.getElementById(`map-container-${viewId}`)!; 
+    try {
+        if (!navigator.onLine) {
+            mapContainer.innerHTML = message;
+            throw new Error("Browser is offline! GeoBlazor requires an internet connection to work with ArcGIS.");
+        }
+        fetch('https://www.arcgis.com/sharing/rest')
```
morehavoc, line 1082: I'm not sure we should do this check. this guarantees that we are "online" however, It wouldn't be hard for someone to use the API behind their firewall (with no outside access) if we _don't_ check this. 

We are already bundling the ArcGIS JS API as needed, so I think the only external resource is this: `<link href="https://js.arcgis.com/4.24/esri/themes/light/main.css" rel="stylesheet"/>`.

If that's true, it's super easy for someone to "host" this themselves and not need "internet" access, only need access to their local network.

Does `navigator.online` correctly respond if you disable the internet in chrome dev tools?

morehavoc (changes requested):
I think we should discuss this a bit more before making the check to `arcgis.com`.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: `navigator.online` was not reliable in my testing, which is why I implemented this.

morehavoc: A few things that we discussed today that we should probably do in this PR:
- Host/deliver *all* of the JS assets (CSS, wasm, etc.) files. [Docs](https://developers.arcgis.com/javascript/latest/es-modules/#working-with-assets)
- Copy asset files on build
- Set the `config.assetsPath` property correctly.
- Optional: Expose other `config` properties that would let you host this on an offline/secured network: `assetsPath` and `fontsUrl` at a minimum, although we should probably expose all of these [config](https://developers.arcgis.com/javascript/latest/api-reference/esri-config.html#assetsPath) properties eventually (if we don't know, let's put them in a gissue for later).

I *think* that if we do all of this, a GeoBlazor app will run "disconnected" meaning that you don't have to have a basemap, and you can add graphics locally and it will work, draw the map etc. So you could run this "offline" although that's not really the purpose, it's really so you can run in an "isolated" network.

morehavoc: Oh, and this means that we shouldn't need to check to see if `arcgis.com` is accessible from the code. If anything, we would check that the `assetsPath` is there, b/c that's what would allow the API itself to work. Any other online sources would just be for data. So we could still do this check, basically see if the `assetsPath` responds, and if it does we are "online" if not, display a message that assets can not be loaded.

@TimPurdum we'll still have to tell users to pick the correct CSS file (they have to pick the theme they want), not sure what that will look like now, but some local path?

Also, for deployment the user will have to ensure that the mime-types are setup on their server correctly ([docs](https://developers.arcgis.com/javascript/latest/install-and-set-up/#web-server-hosting-configuration))

TimPurdum:
```
@morehavoc Thanks for your response, I solved the problem i met already.



```

TimPurdum:
```
  @morehavoc Thanks for your info. Currently, I am still working on the library in "offline" because I just want to analyze the code before loading data(map). I hope it can work in the offline condition someday. I'll let you know if i need something from you. Thanks again.



```

TimPurdum:
@morehavoc , this branch is updated
- packs the css file
- packs the asset files (only if they are not already copied to `wwwroot/assets`)
- assets lookup defaults to `wwwroot/assets`, but can be set in appsettings.json or environment variables.
- created a `Local.razor` test page with no url references
- connectivity check now only checks connection to asset files, updated messages
- adds an `AsyncTask` in MSBuild to offload file copying tasks to background processes during `DEBUG` builds.
- removed all references to `cmd` and `robocopy`, tested on Ubuntu, in preparation for CI/CD work later.

samples/dymaptic.GeoBlazor.Core.Sample.Server/Pages/_Layout.cshtml
```diff
@@ -16,7 +16,6 @@
         <link href="_content/dymaptic.GeoBlazor.Core.Sample.Shared" />
         <script src="_content/dymaptic.GeoBlazor.Core.Sample.Shared/functions.js"></script>
         <component type="typeof(HeadOutlet)" render-mode="ServerPrerendered" />
-        <link rel="stylesheet" href="https://js.arcgis.com/4.24/esri/themes/light/main.css">
```
morehavoc, line 19: Don't we still need to load this CSS file? How else would the person set the theme?

morehavoc: I see, later we are defaulting to the `dark` theme in `core.css`. I think we need a way for the user to control which CSS theme is loaded.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -1066,4 +1071,28 @@ function buildDotNetListItem(item: ListItem): DotNetListItem | null {
...
         actionSections: item.actionsSections
     } as DotNetListItem;
+}
+
+function checkConnectivity(viewId) {
+    let connectError = new Error('Cannot load ArcGIS Assets!');
+    let message = '<div><h1>Cannot retrieve ArcGIS asset files.</h1><p><a target="_blank" href="https://docs/geoblazor.com/assetFiles"</p></div>';
+    let mapContainer = document.getElementById(`map-container-${viewId}`)!; 
+    try {
+        fetch(esriConfig.assetsPath)
```
morehavoc, line 1081: This seems fine, but are there any issues with calling `fetch` like this if the `assetsPath` is a directory? For example, if the server has list directory contents as false, will this fail? Maybe we should check for a specific file within `assetsPath` since we know it *has* to look like an esri assets folder or nothing will work.

docs/pages/assetFiles.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: page
+title: Asset Files
+nav_order: 4
+---
+Asset files, by default, will load from the `dymaptic.GeoBlazor.Core` library at `./_content/dymaptic.GeoBlazor.Core/assets`.
+
+If for some reason you want to bundle your own assets, you would change this by adding an environment variable or setting
+in `appsettings.json` of `"ArcGISAssetsPath": "...your/web/assets/url"`
```
morehavoc, line 9: Some more text for you:

Note that this `ArcGISAssetsPath` can also be a full URL if your assets are hosted on your Enterprise portal, for example. Most people won't need to change this, but here are the [docs](https://developers.arcgis.com/javascript/latest/api-reference/esri-config.html#assetsPath) for this property.

morehavoc (changes requested):

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: Added a path to an actual file

samples/dymaptic.GeoBlazor.Core.Sample.Server/Pages/_Layout.cshtml
TimPurdum: I put these back into link tags, after trying several other approaches, but with a local path to the bundled css

morehavoc (commented):
Are the changes to using the local path for `css` files, etc. breaking? I mean, if a person was already using the library and was getting the CSS from the internet and the assets are coming from the internet, that's still fine, right?

TimPurdum:
> Are the changes to using the local path for css files, etc. breaking? I mean, if a person was already using the library and was getting the CSS from the internet and the assets are coming from the internet, that's still fine, right?

Yes, referencing the online versions of the css or even js shouldn't matter. The JS is going to be ignored once the `assetsPath` is set, and CSS just ignores duplicate definitions, I believe. I can test that once in the morning to be sure.

TimPurdum:
@morehavoc I'd like to wrap up this PR today if possible.

morehavoc (approved):

TimPurdum:
I tested multiple combinations of CSS links, no way to really "break" it! Also updated the css links in `gettingStarted.md`.

PR #96, opened 2022-10-23
Feature Layer graphics

### LINKED ISSUE #92: Support Graphics added to FeatureLayer
According to the [Feature Layer](https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-FeatureLayer.html) JS docs, a `Feature` in ArcGIS JS is just a `Graphic`.

- Add the ability to add and register Graphics as subcomponents of `FeatureLayer`
- Alter add, remove, and update logic for graphics to make sure it handles feature layer graphics as well as `GraphicLayer` graphics and those placed directly on `MapView`.

TimPurdum:
Closes #92 

- Added the ability to use `Graphics` as local, client-side features in a `FeatureLayer`
- Added `Field` for `FeatureLayer` attributes
- Added parameterized constructors for the following classes to avoid BL0005 warning (setting razor Parameters outside razor)
   - Extent
   - Point
   - PolyLine
   - Polygon
   - SpatialReference
   - Graphic
   - FieldInfo
   - FieldInfoFormat
   - FieldsPopupContent
   - PopupTemplate
   - PopupOptions
- Added convenience `Add` and `Remove` methods to `FeatureLayer`, `GraphicsLayer`, and `MapComponent` that call `RegisterChildComponent` (or `UnRegister...` for `Remove`) under the hood.
- Added to the `FeatureLayers.razor` sample page with an example of setting a feature layer from local graphics added in C#


TimPurdum:
@morehavoc can you review this PR today?

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/BasemapProjections.razor
```diff
@@ -41,5 +41,5 @@
         return Task.CompletedTask;
     }
 
-    private SpatialReference _spatialReference = new() { Wkid = 102100 };
+    private SpatialReference _spatialReference = new(102100);
```
morehavoc, line 44: Does us changing these to have constructors constitute a breaking change?

src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs
```diff
@@ -11,20 +11,6 @@ namespace dymaptic.GeoBlazor.Core.Components.Geometries;
 [JsonConverter(typeof(GeometryConverter))]
 public class Geometry : MapComponent
 {
-    /// <summary>
```
morehavoc, line 14: How come these are going away?

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -160,7 +158,7 @@ export async function buildMapView(id: string, dotNetReference: any, long: numbe
                 });
                 let center;
                 let spatialRef;
-                if (spatialReference !== undefined && spatialReference !== null) {
+                if (hasValue(spatialReference)) {
```
morehavoc, line 161: Nice!

morehavoc (approved):

codergrl:
@TimPurdum I got the chance to test the sample and then got pulled away on client calls, but while testing I noticed a bunch of errors you should probably look into before committing this. I basically just clicked to turn the layers on and off randomly in the sample. 

image

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/BasemapProjections.razor
TimPurdum: No, you can still call the parameterless constructor, but you will get a warning (which is how it was previously).

src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs
TimPurdum: As we discussed on Discord, these were non-functional, so should not have been included. Setting them did nothing, and they would always return `null`. I will see if I can implement their real functionality and add them back in soon.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: you can only write the same thing so many times.... 🙂

PR #97, opened 2022-10-23
Implemented GeometryEngine

### LINKED ISSUE #75: Implement full GeometryEngine
https://developers.arcgis.com/javascript/latest/api-reference/esri-geometry-geometryEngine.html

1. Create a passthrough C# `LogicComponent` that exposes all methods for the GeometryEngine.
2. Create a sample page to demonstrate some of the functionality.

TimPurdum:
Closes #75 

- Added `GeometryEngine.cs` and `GeometryEngine.ts`
- Moved sample `CalculateGeometries.razor` from interacive
- Added more type conversion helpers from .NET to JS
- Created DI extension `.AddGeoBlazor` to import logic components (GeometryEngine, Projection).
- Cleaned up some TS references


seahro (approved):
new documentation files, adjusted sample file names and reference to files and corresponding methods including the geometryEngine.  some method exporting also.

src/dymaptic.GeoBlazor.Core/DependencyExtension.cs
```diff
@@ -0,0 +1,21 @@
...
+/// </summary>
+public static class DependencyExtension
+{
+    /// <summary>
+    ///     Adds the Logic components <see cref="GeometryEngine"/> and <see cref="Projection"/> to your dependency
+    ///     injection collection.
+    /// </summary>
+    public static IServiceCollection AddGeoBlazor(this IServiceCollection serviceCollection)
+    {
+        return serviceCollection.AddScoped<GeometryEngine>()
```
morehavoc, line 18: Is scoped the right level here? These are static in the JS api.

src/dymaptic.GeoBlazor.Core/Model/GeometryEngine.cs
```diff
@@ -0,0 +1,896 @@
...
+using dymaptic.GeoBlazor.Core.Objects;
+using Microsoft.Extensions.Configuration;
+using Microsoft.JSInterop;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+
+namespace dymaptic.GeoBlazor.Core.Model;
+
+/// <summary>
```
morehavoc, line 11: These look great! Do any of them deviate from the JS API?

src/dymaptic.GeoBlazor.Core/Objects/NearestPointResult.cs
```diff
@@ -0,0 +1,31 @@
+﻿using dymaptic.GeoBlazor.Core.Components.Geometries;
+using dymaptic.GeoBlazor.Core.Model;
+
+
+namespace dymaptic.GeoBlazor.Core.Objects;
+
+/// <summary>
+///     Object returned from the nearestCoordinate(), nearestVertex(), and nearestVertices() methods of <see cref="GeometryEngine"/>.
+/// </summary>
+public class NearestPointResult
```
morehavoc, line 10: I noticed that Geometries seem to be missing the `M` property while looking at this return.

Also, why are Geometries in `Components` not in `Model`?

I looks like the JS geometry engine has not yet caught up to the Runtime engine and doesn't have all of the linear referencing functions, but we should still track and provide the `M` values.

src/dymaptic.GeoBlazor.Core/DependencyExtension.cs
TimPurdum: Good point, I think I'll change these to Singletons to match

src/dymaptic.GeoBlazor.Core/Model/GeometryEngine.cs
TimPurdum: Not really, no. They're very much 1-1 with the exception of having to write out parameter overrides.

src/dymaptic.GeoBlazor.Core/Objects/NearestPointResult.cs
TimPurdum: Geometries is in `Components` because it can be declared as a razor component in markup.
I'm adding `M` to `Point`, I don't see that as a property on the other geometries in the JS API docs. Some of them have `hasM`, but that's calculated and shouldn't be a settable field as far as I can tell.

src/dymaptic.GeoBlazor.Core/DependencyExtension.cs
TimPurdum: Testing on develop (should have tested before merging), I have to put these back. `JSRuntime` is scoped, so these have to be as well, or it throws an exception. I think that's fine, however, because the scope encompasses a page and internal navigation until a user or code refreshes the browser.

PR #101, opened 2022-10-30
feature/90_return_geometry_from_view_click_event - initial PR to issue 90

Sipowiz:
- Added `GeometrySelection` page to samples project
- Added link in `NavMenu `to `GeometrySelection` page (missing proper icon)
- Modified `Geometry` base component to have required members required for feature
- Added new delegate to be used in events that can have async handler (will be moved eventually to own file)
- Added new public event to `MapView` which fires when view gets rendered
- Added event args associated with geometry selected event
- Added `geometry.ts` that holds the js/ts logic for handling geometry selectioning in ArcGis

### DIFF
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/GeometrySelection.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/GeometrySelection.razor
@@ -1,0 +1,80 @@
+﻿@page "/geometry-selection"
+@using dymaptic.GeoBlazor.Core.Model.EventArgs
+
+<h3>GraphicSelection</h3>
+<MapView Longitude="_longitude"
+         Latitude="_latitude"
+         Style="height: 600px; width: 100%;">
+    <Map ArcGISDefaultBasemap="arcgis-topographic">
+        <GraphicsLayer>
+            @foreach (var feature in Features)
+            {
+                <Graphic Attributes="feature.Attributes">
+                    <Point OnGeometrySelected="GeometrySelected" Longitude="feature?.Point?[0]" Latitude="feature?.Point?[1]" />
+                    <SimpleMarkerSymbol Color="@(new MapColor(81, 46, 132))">
+                        <Outline Color="@(new MapColor(255, 255, 255))" Width="1" />
+                    </SimpleMarkerSymbol>
+                </Graphic>
+            }
+        </GraphicsLayer>
+    </Map>
+</MapView>
+
+@code {
+
+    double _latitude = 34.027;
+    double _longitude = -118.805;
+
+    List<MyObject> Features { get; set; } = GetFeatures();
+
+    Task GeometrySelected(GeometrySelectedEventArgs<Geometry> e)
+    {
+        return Task.CompletedTask;
+    }
+
+    static List<MyObject> GetFeatures() => 
+        new(){
+            new(){
+                Point = new MapPoint(-118.818984489994, 34.0137559967283),
+                Attributes = new List<KeyValuePair<string, object>>(){
+                    new("address", "Test Street 1"),
+                    new("zipcode", "00530")
+                }.ToDictionary(i => i.Key, i => i.Value)
+            },
+            new(){
+                Point = new MapPoint(-118.806796597377, 34.0215816298725),
+                Attributes = new List<KeyValuePair<string, object>>()
+                {
+                    new("address", "Test Street 2"),
+                    new("zipcode", "00531")
+                }.ToDictionary(i => i.Key, i => i.Value)
+            },
+            new(){
+                Point = new MapPoint(-118.791432890735, 34.0163883241613),
+                Attributes = new List<KeyValuePair<string, object>>(){
+                    new("address", "Test Street 3"),
+                    new("zipcode", "00532")
+                }.ToDictionary(i => i.Key, i => i.Value)
+            },
+            new(){
+                Point = new MapPoint(-118.79596686535, 34.008564864635),
+                Attributes = new List<KeyValuePair<string, object>>(){
+                    new("address", "Test Street 4"),
+                    new("zipcode", "00533")
+                }.ToDictionary(i => i.Key, i => i.Value)
+            },
+            new(){
+                Point = new MapPoint(-118.808558110679, 34.0035027131376),
+                Attributes = new List<KeyValuePair<string, object>>(){
+                    new("address", "Test Street 5"),
+                    new("zipcode", "00534")
+                }.ToDictionary(i => i.Key, i => i.Value)
+            },
+        };
+
+    private class MyObject
+    {
+        public MapPoint? Point { get; set; }
+        public Dictionary<string, object> Attributes { get; set; } = new();
+    }
+}
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Shared/NavMenu.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Shared/NavMenu.razor
@@ -87,7 +87,8 @@
         new("projection-tool", "Projection Tool", "oi-cog"),
         new("basemap-projections", "Basemap Projections", "oi-flash"),
         new("unique-value", "Unique Renderers", "oi-brush"),
-    new("marker-rotation", "Marker Rotation", "oi-loop-circular")
+        new("marker-rotation", "Marker Rotation", "oi-loop-circular"),
+        new("/geometry-selection", "Geometry Selection", "oi-compass"),
     };
 
     private record PageLink(string Href, string Title, string IconClass);
--- a/src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs
@@ -1,6 +1,9 @@
-﻿using System.Text.Json;
-using System.Text.Json.Serialization;
+﻿using dymaptic.GeoBlazor.Core.Model.EventArgs;
 using Microsoft.AspNetCore.Components;
+using Microsoft.Extensions.Logging;
+using Microsoft.JSInterop;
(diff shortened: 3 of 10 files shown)


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/GeometrySelection.razor
```diff
@@ -0,0 +1,80 @@
...
+            new(){
+                Point = new MapPoint(-118.808558110679, 34.0035027131376),
+                Attributes = new List<KeyValuePair<string, object>>(){
+                    new("address", "Test Street 5"),
+                    new("zipcode", "00534")
+                }.ToDictionary(i => i.Key, i => i.Value)
+            },
+        };
+
+    private class MyObject
```
TimPurdum, line 75: This could use a more descriptive name. Also might be a good candidate for making a `record` since the data is immutable (never updated).

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Shared/NavMenu.razor
```diff
@@ -87,7 +87,8 @@
         new("projection-tool", "Projection Tool", "oi-cog"),
         new("basemap-projections", "Basemap Projections", "oi-flash"),
         new("unique-value", "Unique Renderers", "oi-brush"),
-    new("marker-rotation", "Marker Rotation", "oi-loop-circular")
+        new("marker-rotation", "Marker Rotation", "oi-loop-circular"),
+        new("/geometry-selection", "Geometry Selection", "oi-compass"),
```
TimPurdum, line 91: can you find a different openiconic icon for this new page? We're already using `oi-compass` for the `Navigation` page.

Sipowiz:
Okay! Thanks for your comments @TimPurdum. I edited title and description information. Whole Forking thing is new to me so didn't realize where all of that information gets linked.

src/dymaptic.GeoBlazor.Core/Scripts/geometry.ts
```diff
@@ -0,0 +1,47 @@
+﻿import { arcGisObjectRefs } from "./arcGisJsInterop";
+let dotNetRef: any = null;
+
+export function initialize(dotNetReference: any, viewId: string, geometryId: string, geometryType: string): void {
```
TimPurdum, line 4: A few thoughts here.
- Unless I'm missing something, your main logic is commented out?
- We already have a listener to click events on the view, and more explicit ones coming in #102. I think @morehavoc's suggestion was to expose `hitTest` in the C# as a separate message on `MapView`. That way, you would chain in C#: in a click event callback, pass the event to `view.HitTest(point)` and get a response. Now, I don't mind us wrapping up that logic _as well_ as exposing it, but we should expose it first, then write a convenience method around it, something like `public List<Geometry> OnClickGeometries` in `MapView`.
- In general, I am aware that `MapView.razor.cs` and `arcGisObjectRefs.ts` are getting cumbersome and large, so I'm not opposed to adding new `.ts` files and spreading out the logic. But as you discovered when fighting with imports, it does have a cost associated.
- If there's a good use case for each geometry to track their own selection, rather than returning a list from the view, let's keep going with this. However, like I said above, I think we need to implement a view-level response first, and then maybe have that response trigger the geometry-level response?

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/GeometrySelection.razor
```diff
@@ -0,0 +1,80 @@
...
+@code {
+
+    double _latitude = 34.027;
+    double _longitude = -118.805;
+
+    List<MyObject> Features { get; set; } = GetFeatures();
+
+    Task GeometrySelected(GeometrySelectedEventArgs<Geometry> e)
+    {
+        return Task.CompletedTask;
```
TimPurdum, line 32: What is the ultimate use case for the samples page? Will we display something that says "Geometry x was clicked!"?

src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs
```diff
@@ -18,6 +21,12 @@ public class Geometry : MapComponent
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public bool? HasM { get; set; }
 
+    /// <summary>
+    /// This callback gets fired when this geometry gets selected
+    /// </summary>
+    [Parameter]
+    public EventCallback<GeometrySelectedEventArgs<Geometry>> OnGeometrySelected { get; set; }
```
TimPurdum, line 28: nice use of default blazor patterns!

Sipowiz:
I'm used to working in a repo that isn't forked but rather every feature or bug gets it's own branch when starting to work on them and where commits are pushed to that branch frequently and once feature is done / bug is fixed and it's tested, whole branch gets merged to staging and afterwards to production. Admitted these projects haven't had any other contributors besides myself and they are not libraries but full blown SaaS services. Exited to be contributing to this.

src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs
```diff
@@ -42,6 +51,44 @@ public class Geometry : MapComponent
...
+    /// Bool to determine if geometry JS module has been initialized
+    /// </summary>
+    bool ModuleInitialized { get; set; }
+
+    /// <inheritdoc />
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        if (firstRender)
+        {
+            _geometryJSModule = JsRuntime?.InvokeAsync<IJSObjectReference>("import", $"./_content/dymaptic.GeoBlazor.Core/js/geometry.js").AsTask();
```
TimPurdum, line 75: As you probably ran into, `OnAfterRender` in Blazor does _not_ mean that the ArcGIS map has rendered yet. Hence your check for `MapRendered` below. In fact, I'm not sure whether `View` is populated here either. You might have to wait for _not_ `firstRender` to hook this up, or try moving line 79 (`await base.OnAfterRenderAsync`) up, since that will do the registration of this component with parent components.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -14,6 +14,13 @@
 
 namespace dymaptic.GeoBlazor.Core.Components.Views;
 
+/// <summary>
+/// Delegate Task that can be used asyncronously
+/// </summary>
+/// <param name="sender"></param>
```
TimPurdum, line 20: I see why you couldn't use Blazor's `EventCallback` for this event, since it's being listened for in a child component. However, a few things:
- I would move this to the bottom of the file, or to its own file
- Is the `sender` parameter necessary? Unless we're going to use this pattern outside of `MapView`, the sender is obvious and already known.
- Go ahead and delete xml comment lines for `param` and `returns` if they are obvious. Our docs site (docs.geoblazor.com) is partially generated from XML comments, so we don't want to leave empty ones around.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -24,6 +31,11 @@ namespace dymaptic.GeoBlazor.Core.Components.Views;
 /// </example>
 public partial class MapView : MapComponent
 {
+    /// <summary>
+    /// Event fires allways when view has rendered
+    /// </summary>
+    public event AsyncEventHandler? ViewRendered;
```
TimPurdum, line 37: Assuming the point of this event is for internal components to receive notification, I think we could make this `internal` instead of public, to avoid confusion for consumers with `OnMapRenderedHandler`. We could also name it something special like `ViewRenderedInternalEvent`. Also, curious on your thoughts if my `OnMaprenderedHandler` and my PR's `OnMapRendered` EventCallback should actually be `OnViewRendered`? Since views and maps are 1-1, I tend to conflate them more than I should.

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
```diff
@@ -79,9 +81,9 @@
         <Exec Command="npm run releaseBuild" />
     </Target>
     
-    <Target Name="Copy Documentation" AfterTargets="Build">
+    <!--<Target Name="Copy Documentation" AfterTargets="Build">
```
TimPurdum, line 84: Make sure to clean up this csproj file as well. Lines 51 and 63-66 look suspect to me.

TimPurdum (changes requested):
OK, I'm done reviewing! There's a lot of nice work here, and I'm impressed with how well you picked up the project structure and approach. I think the main thing going forward is let's figure out `HitTest` and a view-first implementation, and then we can build off that.

TimPurdum:
> I'm used to working in a repo that isn't forked but rather every feature or bug gets it's own branch when starting to work on them and where commits are pushed to that branch frequently and once feature is done / bug is fixed and it's tested, whole branch gets merged to staging and afterwards to production. Admitted these projects haven't had any other contributors besides myself and they are not libraries but full blown SaaS services. Exited to be contributing to this.

The pattern should be similar here, except we're using your branch in _your_ repo as the feature branch. Once we're done tweaking, testing, etc., we will merge it into out repo's `develop` branch.

TimPurdum:
@Sipowiz how are you feeling about this PR? I could take it over in the next few weeks and wrap up implementing if you want, but don't want to step on your toes if you're still interested in finishing it.

Sipowiz:
> @Sipowiz how are you feeling about this PR? I could take it over in the next few weeks and wrap up implementing if you want, but don't want to step on your toes if you're still interested in finishing it.

@TimPurdum i've been really busy with my day job, so haven't had the time to finish this. Looks like it'll be like that for at least the rest of the year so i'm more than willing to hand this over to you.

PR #102, opened 2022-10-30
Feature/91 event handlers

### LINKED ISSUE #91: Add OnViewChanged Event Handler
Brought up during [this Discord discussion](https://discord.com/channels/1027907220949717033/1031452704167432222). User `Sipowicz` offered some sample implementation.

Add event listeners in `arcGisInterop.ts` that track any user input that would change the map extent (e.g. `mouse-wheel`, `drag`). Pass these back to `MapView` in a consolidated event, something like `OnExtentChanged`, with the new `Extent`. Expose as a handler for consuming code to listen.

TimPurdum:
Added Event Handlers and ReactiveUtils
Closes #91 
- All JS-style regular events are exposed as `EventCallback`s in `MapView` with simplified return types (e.g., just the `Point`)
- Reactive methods added to `MapComponent` to support calling on different component types (layer, widget, view)
- Samples page in UI
- new docs page explaining the options


src/dymaptic.GeoBlazor.Core/Components/MapComponent.razor.cs
```diff
@@ -240,6 +271,421 @@ protected virtual async Task RenderView(bool forceRender = false)
...
+    ///     The type of return value to expect in the handler.
+    /// </typeparam>
+    /// <exception cref="UnMatchedTargetNameException">
+    ///     This exception is thrown when a watchExpression's target object name doesn't match the targetName parameter.
+    /// </exception>
+    /// <remarks>
+    ///     For adding watchers to types other than <see cref="MapView"/> and <see cref="SceneView"/>, the default <code>targetName</code> should not be relied upon. Make sure it matches the variable in your <code>watchExpression</code>
+    /// </remarks>
+    // ReSharper disable once UnusedMethodReturnValue.Global
+    public Task AddReactiveWatcher<T>(string watchExpression, Action<T> handler, string? targetName = null,
```
morehavoc, line 351: We talked about this the other day, but I'm worried about how to know what my options are for this `T`. Just knowing some basic things like how we map classes from the JS namespace to the GeoBlazor namespace would help in the events doc.

If I recall, using `<string>` will cause it to return the JSON string of the object to use myself, which is great! Are there any conflicts between that JSON representation and something that actually returns a string?

src/dymaptic.GeoBlazor.Core/Components/MapComponent.razor.cs
```diff
@@ -240,6 +271,421 @@ protected virtual async Task RenderView(bool forceRender = false)
...
+        {
+            throw new UnMatchedTargetNameException(targetName, watchExpression);
+        }
+        IJSObjectReference? jsRef = 
+            await JsModule!.InvokeAsync<IJSObjectReference?>("addReactiveWatcher", Id, 
+                targetName, watchExpression, once, initial);
+
+        if (jsRef != null)
+        {
+            _watchers[watchExpression] = (handler, jsRef);
```
morehavoc, line 373: This is limiting to one handler per watchExpression?

src/dymaptic.GeoBlazor.Core/Components/MapComponent.razor.cs
```diff
@@ -240,6 +271,421 @@ protected virtual async Task RenderView(bool forceRender = false)
...
+                _ => typedValue
+            };
+        }
+
+        handler.DynamicInvoke(typedValue);
+    }
+
+    
+    /// <summary>
+    ///     Tracks any properties accessed in the <code>listenExpression</code> and calls the callback when any of them change.
```
morehavoc, line 432: what's the difference between Listener and Watcher? (is this `on` vs `watch` in JS?)

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -143,87 +145,390 @@ public double Rotation
...
+    }
+    
+    /// <summary>
+    ///     Handler delegate for click events on the view. Must take in a <see cref="Point"/> and return a <see cref="Task"/>.
+    /// </summary>
+    /// <remarks>
+    ///     <b style="color: red">OBSOLETE: Use OnClick EventCallback instead</b>
+    /// </remarks>
+    [Parameter]
+    [Obsolete("Use OnClick EventCallback instead.")]
```
morehavoc, line 212: Should we add a note or date here to know when we can actually pull this code out of the codebase?

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -143,87 +145,390 @@ public double Rotation
...
     public Func<Point, Task>? OnClickAsyncHandler { get; set; }
+    
+    /// <summary>
+    ///     Handler delegate for click events on the view.
+    /// </summary>
+    [Parameter]
+    public EventCallback<Point> OnClick { get; set; }
```
morehavoc, line 219: In general, I think these events should return the same data that the ArcGIS even returns. For example, HitTest returns a full mouse event with additional properties attached. This way you can do things like check for modifier keys being pressed, etc.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -232,11 +537,141 @@ public void OnViewRendered()
...
+    }
+    
+    /// <summary>
+    ///     Fires after a LayerView is destroyed and is no longer rendered in the view. This happens for example when a layer is removed from the map of the view.
+    /// </summary>
+    [Parameter]
+    public EventCallback<LayerViewCreateErrorResult> OnLayerViewCreateError { get; set; }
+    
+    /// <summary>
+    ///     Set this parameter to limit the rate at which recurring events (pointer-move, drag, etc.) are returned.
```
morehavoc, line 668: What events does this apply to?

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -218,25 +221,222 @@ export async function buildMapView(id: string, dotNetReference: any, long: numbe
...
         });
+    }
 
+    if (activeEventHandlers.includes('OnPointerUp')) {
+        view.on('pointer-up', (evt) => {
+            let point = (view as MapView).toMap({
+                x: evt.x,
+                y: evt.y
+            });
+            dotNetRef.invokeMethodAsync('OnJavascriptPointerDown', buildDotNetPoint(point));
```
morehavoc, line 351: Should this be `OnJavascriptPointerUp`?

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -1066,4 +1274,61 @@ function buildDotNetListItem(item: ListItem): DotNetListItem | null {
...
         actionSections: item.actionsSections
     } as DotNetListItem;
+}
+
+
+export function addReactiveWatcher(targetId: string, targetName: string, watchExpression: string, once: boolean, 
+                                   initial: boolean) : any {
+    let target = arcGisObjectRefs[targetId];
+    let dotNetRef = dotNetRefs[targetId];
+    console.log(`Adding watch: "${watchExpression}"`);
```
morehavoc, line 1284: if we make these `debug` will they go away in the compiled JS? Even if not, that might be nice to do at the debug level to reduce the amount of logging that is visible by default (log).

morehavoc (changes requested):
This is great, as predicted I still think that we should return the full event objects (not just the point), especially for mouse and keyboard events where modifier keys can be very helpful.

src/dymaptic.GeoBlazor.Core/Components/MapComponent.razor.cs
TimPurdum: I don't think there is a conflict, string will handle both raw strings and json. I agree about the docs. After I fix the other event handlers to return event types, we can document all the possibilities for each event, or at least the most common.

src/dymaptic.GeoBlazor.Core/Components/MapComponent.razor.cs
TimPurdum: That's correct. I can't imagine wanting to hook up multiple, when they would fire simultaneously, as opposed to having a single handler that did multiple things.

src/dymaptic.GeoBlazor.Core/Components/MapComponent.razor.cs
TimPurdum: Yes, exactly. On/Listener is for events, Watch is for property values.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: My assumption would be when we hit v2.0.0. What did you have in mind for a note?

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: Agreed, I will go back through and update the returns to all be full event objects.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: Good question. It applies to all "rapidly-firing" events: pointer-move, drag, mouse-wheel, resize, and extent-change. I will add this to the xml.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: nice catch!

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
morehavoc: maybe just a `//TODO: ` tag in a comment?

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: I didn't even know that was a thing in JavaScript! Agreed, I'll play around with it.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -218,26 +221,178 @@ export async function buildMapView(id: string, dotNetReference: any, long: numbe
...
 
+        setEventListeners(view, dotNetRef, eventRateLimitInMilliseconds, activeEventHandlers);
+        
+        unsetWaitCursor(id);
+    } catch (error) {
+        logError(error, id);
+    }
+}
+
+function setEventListeners(view: __esri.View, dotNetRef: any, eventRateLimit: number | null,
```
morehavoc, line 232: Does this only get called when the mapview is created? If yes, we are not supporting the scenario where you might connect and disconnect to events at different points in the app?

morehavoc (approved):

docs/pages/classes/dymaptic.GeoBlazor.Core.Events.BlurEvent.md
```diff
@@ -0,0 +1,76 @@
+---
+layout: default
+title: BlurEvent
+parent: Classes
+---
+#### [dymaptic.GeoBlazor.Core](index.html 'index')
```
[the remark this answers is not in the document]
TimPurdum: hmm, good catch. I think there's no "cleanup" in the processing, so when something gets moved, it leaves docs behind. I'll fix that.

docs/pages/reactive.md
```diff
@@ -0,0 +1,128 @@
+---
```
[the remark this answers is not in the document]
TimPurdum: I don't think this is true for the ReactiveUtils events. You could add as many as you want. For `EventCallback` events from `MapView`, that's part of the Blazor framework, that they are parameters and thus only consumed once. But I can document that.

PR #103, opened 2022-11-04
Feature/63 add georss layer

codergrl:
Adds GeoRSS Layer and corresponding sample.


.gitignore
```diff
@@ -369,3 +369,5 @@ MigrationBackup/
 /docs/vendor
 /docs/.DS_Store
 /src/dymaptic.GeoBlazor.Core/Documentation/**/*
+
+package-lock.json
```
TimPurdum, line 373: Thanks for adding this!

package.json
```diff
@@ -0,0 +1,5 @@
+{
```
TimPurdum, line 1: Not sure where this one came from, but we shouldn't need this in the root folder. We've already got `src/dymaptic.GeoBlazor.Core/package.json`

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
```diff
@@ -36,6 +36,7 @@
         <PackageReference Include="DefaultDocumentation" Version="0.8.2" PrivateAssets="all" />
         <PackageReference Include="Microsoft.AspNetCore.Components.Web" Version="6.0.6" />
         <PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.1" />
+        <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="5.0.0" />
```
TimPurdum, line 39: Since user secrets is actually part of the executable apps (e.g., `Samples.Server`), this line should be removed.

src/dymaptic.GeoBlazor.Core/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
```
TimPurdum, line 1: Since this file was added to .gitignore, we actually also need to _delete_ it and commit, to make it disappear from the repo.

TimPurdum (approved):
Nice work! Just a few things to clean up, then you can merge this.

.gitignore
morehavoc: I believe that package lock is supposed to be committed to the repo. It keeps the resolved npm version numbers.

codergrl: ok, adding it back and removing it from .gitignore

TimPurdum: @morehavoc it's definitely a build-generated product, so it should be recreated with no issues if left out.

morehavoc: @TimPurdum The `package-lock.json` file can be generated by the build, but it can also be consumed by the build. The `package-lock.json` contains the _exact_ versions of the packages that were used in the build, while the `package.json` file contains only the relative, hopefully semantically versioned package versions. 

The internet seems pretty torn on this (at least in some places). But the general consensus is that if you are building an NPM package you should *not* commit `package-lock.json` because it is ignored by the npm package manager when resolving dependencies for each library. But for building an app, which is what we have in terms of JS (it not a NPM module, it is an app) you should commit `package-lock.json`.

In my experience `package-lock.json` doesn't change unless you change something in `package.json` or something happened and a version of the package that was specified in package-lock doesn't exist on NPM any longer, which does happen sometimes.

Issue #109, opened 2022-11-25
CSS for dymaptic branding and map-wrapper not loaded, doesn't support fixed width


TimPurdum:
Discovered while implementing a new sample app. The following line needs re-added to the `<head>` tags of projects to get our bundled css, unless the _consuming project_ also bundles it's own css.

Thus, each project needs either:
`<link href="_content/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.bundle.scp.css" rel="stylesheet" />`
or
`<link href="SampleApp.styles.css" rel="stylesheet" />`
in the head tag.

I also discovered that even when loaded, the dymaptic branding was not positioned correctly if the `MapView` is set to a fixed pixel width, instead of `100%`.

PR #110, opened 2022-11-25
Fix for css fixed-width, update docs

TimPurdum:
Small fix
- rearranged styling for `map-wrapper` and `map-container` so that both fixed and percent-based widths lay out correctly
- added reference to bundled css in `gettingStarted.md` required to load GeoBlazor css correctly (one of the new empty templates from MS has this line commented out)


TimPurdum:
Closes #109

seahro (approved):
looks pretty straight forward, minor tweak of the css

PR #113, opened 2022-12-01
OpenStreetMapLayer

TimPurdum:
@morehavoc while implementing the OSM layer, I was adding properties that were missing to related types, and realized that I had currently coded `MapView`'s `Latitude`, `Longitude`, and `Zoom` as non-nullable. This causes problems if 1) I try to apply them to a 3d scene that also has camera settings, or 2) I try to add an `Extent` to the view.

The solution I've come up with thus far is to make them nullable, which is a minor breaking change, but to add the default values for 2d maps in the typescript layer, since those were the only places previously using that value (I think, still testing).

Anyways, I'm open to suggestions. I don't want to make breaking changes willy-nilly, nor do I really want to move our version # to 2.0.0 yet, but these original-sin errors seem important to fix.


morehavoc (approved):
Setting the defaults for these if they are not specified seems okay, but let's make sure those defaults are called out in the documentation. How are they handled in the esri JS API? if they are not specified, what happens? If they are nullable at that level (because you specify either a zoom/lat/long or an extent) then setting them to nullable out our level (without even the defaults) brings it more in line with the esri API. At least for initial values.  There's some value in letting the Esri API handle error conditions like this though.

The defaults could be confusing b/c it would cause the map to zoom somewhere strange unless the defaults are full world extent.

TimPurdum:
> How are they handled in the esri JS API? if they are not specified, what happens? If they are nullable at that level (because you specify either a zoom/lat/long or an extent) then setting them to nullable out our level (without even the defaults) brings it more in line with the esri API.

Yes, the ArcGIS API assumes everything is nullable. The map or layer's `Extent` is usually the fallback for the zoom/location, and if that also is left off, I believe it just reverts to global scale.

 Given the two very large feature sets I have implemented since this branch started, I am thinking more and more that we should make a breaking-change release in the near future and set these to nullable without any defaults set.

PR #114, opened 2022-12-31
Feature/90 hit test

TimPurdum:
image

- Updated the `LayerCreated` event to return not just the `LayerView` but also `IJsObjectReference` refs for `Layer` and `LayerView`, so that further methods can be called directly on these refs.
- Created `MapView.HitTest` methods that take a `ClickEvent`, `PointerEvent`, or just `Point` and returns a `HitTestResult` with `Graphics` that include `Attributes` fields. Note that there are other types of "Hits" in ArcGIS, so this should be expanded later.
  - Due to the size of graphic data returning, I discovered that we needed to "chunk" the data in Blazor Server, because there is a limit of 32k on SignalR connections (https://github.com/dotnet/aspnetcore/issues/23179). This could possibly be refactored into using `Stream` in the future. The other option was to force users to remove the SignalR limit, and that seemed like a bad idea.
- Created new samples page `HitTests.razor`
- Created methods on `LayerView` to support the new sample. 
  - `CreateQuery`
  - `QueryObjectIds`
  - `Highlight`
- greatly expanded `Query.cs` with most of the expected properties.
- fixed a bug with `ArcGISApiKey` in WASM mode.


docs/pages/reactive.md
```diff
@@ -147,3 +147,4 @@ private async Task<bool> AttachZoomWatcher()
     return await _view!.AwaitReactiveSingleWatchUpdate<bool>("view?.zoom > 20");
```
morehavoc, line 147: This is pretty cool! If you did this waiting a lot, would it cause thread exhaustion or something strange?

morehavoc (approved):
I'm excited to have hit test and highlight in the API!

docs/pages/reactive.md
TimPurdum: Yes, I could see this causing exhaustion or even wasm single-thread deadlock if used a lot. But I would think the other methods that are reusable (don't fire just once) are more likely to be heavily used, and those all use a callback pattern which would avoid issues. I would add a warning here, but I'm not _entirely_ sure there is an issue, or how exactly it would manifest.

morehavoc: Sounds good, let’s just see how it goes!

Issue #115, opened 2023-01-01
Add Query Methods to FeatureLayer and FeatureLayerView


TimPurdum:
In #114 we added `IJSObjectReference` for `Layer` and `LayerView`. This opens up the ability to call methods directly on these classes.

1. Refactor `LayerView` and create sub-types for all different types of `Layer`
2. In `FeatureLayer`, implement the following query methods using the `IJSObjectReference`:
   - createQuery
   - queryExtent
   - queryFeatureCount
   - queryFeatures
   - queryObjectIds
   - queryRelatedFeatures
   - queryRelatedFeaturesCount
   - queryTopFeatureCount
   - queryTopFeatures
   - queryTopFeaturesExtent
   - queryTopObjectIds
3. In `FeatureLayerView`, implement the following query methods:
   - createAggregateQuery
   - queryAggregates
   - queryExtent
   - queryFeatureCount
   - queryFeatures

PR #119, opened 2023-01-15
Feature/115 query methods

TimPurdum:
This is only part of #115, although I think now that should be broken up and `relatedFeatures` and `topFeatures` queries should be implemented separately, as should the rest of the `FeatureLayerView` query methods.

1. Implemented a `wrapper` pattern using a JavaScript class for `featureLayer.ts`. This allows a shim layer for converting object types, and also prevents re-bundling the entire ArcGIS package in each file. Instead, the new file is bundled into the main `arcGisJsInterop.js`. I also moved `geometryEngine.ts` and `projection.ts` into this pattern, which cuts down on build time and deployment size significantly.
2. Implemented the basic query methods on `FeatureLayer`
    - createQuery
    - queryExtent
    - queryFeatures
    - queryFeatureCount
    - queryObjectIds
3. Created `AbortManager` to convert .NET `CancellationToken` to JavaScript `AbortController`.
4. Greatly expanded the supporting classes for `Query` and `Popup`, including most popup content types.
5. Implemented `Load` on layers to load and pull data before the view is rendered.
6. Created a generic `EnumToKebabCaseStringConverter` JSON converter to handle all enums.
7. Lots of refactoring to get the new `ServerSideQueries.razor` sample page working, and then to fix any new bugs created in other samples.


seahro (approved):
major rework of the code, whew!

PR #128, opened 2023-02-17
leverage Directory.Build.props

SimonCropp:


TimPurdum (commented):
Hi @SimonCropp! Thanks for the PR. I realize it's pretty self-explanatory, and we don't have established rules for PRs in this repo, but I'd love a quick explanation for anyone who runs into this explaining the advantage of using `Directory.Build.props`. If you can throw in a comment about that, I'll be happy to merge!

SimonCropp:
@TimPurdum Directory.Build.props allows you to have common properties that will then be inherited by all projects in the solution. it allows you to avoid duplicating the same settings in projects, and avoids inconsistency.

TimPurdum (approved):

PR #134, opened 2023-02-24
Feature/86 popup actions

TimPurdum:
Implements `ActionButton` and `ActionToggle` in `PopupWidget` and `PopupTemplate`.

Also bumps beta version to 9, probably last feature before release.


src/dymaptic.GeoBlazor.Core/Components/Layers/Layer.cs
```diff
@@ -170,7 +170,7 @@ public async Task Load(CancellationToken cancellationToken = default)
         IJSObjectReference arcGisJsInterop = await GetArcGisJsInterop();
         JsLayerReference = await arcGisJsInterop.InvokeAsync<IJSObjectReference>("createLayer", 
             // ReSharper disable once RedundantCast
-            cancellationToken, (object)this, true);
+            cancellationToken, (object)this, true, View?.Id);
```
seahro, line 173: in the samples, this would be the mapView id?  Would it be useful to update more of the samples to have mapView ids?...in future releases or could this be more of a suggestion in the documentation.  Something along the lines: as a developer creates more complex mapViews, Layers, etc; implementing some of these classes such as popups will require more elements to have designated Ids for the best customization.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Popups.razor
```diff
@@ -0,0 +1,98 @@
...
+        <Basemap>
+            <PortalItem Id="f35ef07c9ed24020aadd65c8a65d3754" />
+        </Basemap>
+        @if (_showPopupOption == 1)
+        {
+            <FeatureLayer Url="https://services3.arcgis.com/GVgbJbqm8hXASVYi/arcgis/rest/services/Trailheads_Styled/FeatureServer/0"
+                          OutFields="@(new[] { "TRL_NAME", "CITY_JUR", "X_STREET", "PARKING", "ELEV_FT" })">
+                <PopupTemplate Title="Trailhead"
+                               StringContent="<b>Trail:</b> {TRL_NAME}<br><b>City:</b> {CITY_JUR}<br><b>Cross Street:</b> {X_STREET}<br><b>Parking:</b> {PARKING}<br><b>Elevation:</b> {ELEV_FT} ft" />
+            </FeatureLayer>
```
seahro, line 55: So the essential template starts with Field values set in the `FeatureLayer`.  Then we have a formatted content string with both the "values" and corresponding "fields" to display in the popup that is the value for the `PopupTemplate`.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Popups.razor
```diff
@@ -0,0 +1,98 @@
...
+        }
+        else if (_showPopupOption == 2)
+        {
+            <FeatureLayer Url="https://services3.arcgis.com/GVgbJbqm8hXASVYi/arcgis/rest/services/Trails_Styled/FeatureServer/0"
+                          OutFields="@(new[] { "TRL_NAME", "ELEV_GAIN" })">
+                <PopupTemplate Title="Trail Information">
+                    <MediaPopupContent>
+                        <ColumnChartMediaInfo Caption="">
+                            <ChartMediaInfoValue Fields="@(new[] { "ELEV_MIN", "ELEV_MAX" })"
```
seahro, line 64: wow...taking it to 11... charts.

seahro (approved):
Looks good.  I saw the text and chart.  I cant wait to see other media.

src/dymaptic.GeoBlazor.Core/Components/Layers/Layer.cs
TimPurdum: I'm only passing it here because it's needed to register the popup action from popup templates to the view.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Popups.razor
TimPurdum: If you look at the esri docs for `PopupTemplate`, you'll see there are a lot of options around the `Content`. It's basically raw html, Arcade queries, or plain text.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Popups.razor
TimPurdum: This whole page was actually already in our `Pro` library, and I moved it here.

PR #136, opened 2023-02-25
Fix re-render of JS layer bug

### LINKED ISSUE #135: Bug - Round-trip required field validation on layers imported from JS
@seahro reported this bug today, where calling `mapView.Refresh()` would throw some errors that pointed to the fact that it was trying to re-render JS-imported layers.

- Mark incoming JS/Service layers as "Imported"
- Prevent re-rendering of imported layers
- Improve the experience around basemap layers (where this was discovered)

TimPurdum:
Closes #135 , fixing re-render and basemap layer issues

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
+++ b/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
@@ -100,6 +100,7 @@ import GraphicsLayerWrapper from "./graphicsLayer";
 import Popup from "@arcgis/core/widgets/Popup";
 import ElevationLayer from "@arcgis/core/layers/ElevationLayer";
 import PopupWidgetWrapper from "./popupWidgetWrapper";
+import PortalItem from "@arcgis/core/portal/PortalItem";
 
 export let arcGisObjectRefs: Record<string, Accessor> = {};
 export let dotNetRefs = {};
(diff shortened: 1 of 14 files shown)


seahro (approved):
I had one question about the addition to portalItem in arcgisJSInterop and future additions like this.  Looks like minor tweaks to the classes and some targeted adjustments to the actions on the view or the layers, depending on the class.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -100,6 +100,7 @@ import GraphicsLayerWrapper from "./graphicsLayer";
 import Popup from "@arcgis/core/widgets/Popup";
 import ElevationLayer from "@arcgis/core/layers/ElevationLayer";
 import PopupWidgetWrapper from "./popupWidgetWrapper";
+import PortalItem from "@arcgis/core/portal/PortalItem";
```
seahro, line 103: one question.  PortalItem has been operational as part of the mapview up until this last week.  The adjustment to the layer class created a problem with the existing Portalitem reference here, which required a reworking of the code and an import statement here.  Do you see the need to continue to add these imports which are already apart of different classes into the arcGisJSInterop imports as the code continues to evolve?

TimPurdum: This is an import, same as a `using` in C#. In this case, it's unused, I had added it while trying out something, and then deleted the calling code. It's not harmful, but can be removed (I will clean it up). 

The reason it's not necessary is that ArcGIS does a _lot_ of auto-casting objects to defined types. So if you look at the uses of `PortalItem` in this file, they are usually just declared as generic objects like so:
```javascript
{
    id: id
}
```

Issue #151, opened 2023-03-30
Document Issues with MAUI


TimPurdum:
Several issues came up on the Discord channel around setting up and running GeoBlazor in Blazor Hybrid/MAUI. While no code changes are technically necessary, we should do the following:
1. Update the MAUI sample app to show how to load an `appsettings.json` file with an api key
2. Create a new docs page extending `Getting Started` with more details for MAUI

TimPurdum:
This was completed during #145

PR #153, opened 2023-04-02
Feature/145 protobuf streaming

### LINKED ISSUE #145: Investigate ProtoBuf and Streaming to improve graphic serialization
We have discussed many ways of improving the serialization and loading of large data sets
- Use Protobuf serialization instead of JSON (smaller, faster)
- Use `DotNetStreamReference` to pass data as a stream to JavaScript, and read graphics as they come in
- Somehow gain access to the shared memory space between WASM and JS (for wasm only), and pass a pointer to the data
- Combine several of the above, and abstract so that Blazor Server, WASM, and MAUI all work efficiently with the current API methods

TimPurdum:
Closes #145 

This is a large step forward in serializing graphics to JavaScript.
- On the C# side we use the `protobuf-net` nuget package to annotate our `GraphicSerializationRecord` type and serialize it to a `DotNetStreamReference`
- On the JS side we use `protobuf.js` npm package and a JSON type definition file to deserialize the graphics
- All `AddGraphics` calls are chunked, and there is a default and user-controllable chunk size.
- The logic is branched on the Blazor model
   - WASM in .NET 7 uses the synchronous `JSImport` methods instead of async calls
   - Server and MAUI launch multiple tasks for parallel work


TimPurdum:
Also in this PR, I moved all the `Parent.RegisterChildComponent` registrations to earlier in the Blazor life cycle, which avoids this happening on every render. There were some features that depended on this happening after render, however, such as registering layers and graphics from Markup, so you will see a new pattern in those places that register first, then render on the next `OnAfterRenderAsync`.

samples/dymaptic.GeoBlazor.Core.Sample.Maui/MauiProgram.cs
```diff
@@ -27,16 +27,16 @@ public static MauiApp CreateMauiApp()
         builder.Services.AddScoped<HttpClient>();
         builder.Services.AddGeoBlazor();
 
-        string? apiKey = Preferences.Get("ArcGISApiKey", null);
+        builder.Configuration.AddInMemoryCollection();
+
+        var a = Assembly.GetExecutingAssembly();
```
AndersenBell, line 32: you may want to add a comment on why you are doing it this way, what needs to be in the secret (or where to find our documentation on it), and also re-name the variable from "a"

TimPurdum: Good point. I stole this from testing from #151 , but I will probably just merge that work in here, since it is small.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/SerializationTests.razor
```diff
@@ -0,0 +1,75 @@
...
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            PropertyNameCaseInsensitive = true
+        };
+        long unixTimeMilliseconds = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+        MemoryStream ms = new MemoryStream(JsonSerializer.SerializeToUtf8Bytes(_graphics, options));
+        DotNetStreamReference streamRef = new(ms);
+        await _arcGisJsInterop!.InvokeVoidAsync("getFromJsonStream", streamRef, unixTimeMilliseconds);
+    }
+
+    private async Task SendProtobufStream()
```
AndersenBell, line 45: does this need to exist if its commented out?

samples/dymaptic.GeoBlazor.Core.Sample.Wasm/dymaptic.GeoBlazor.Core.Sample.Wasm.csproj
```diff
@@ -1,21 +1,21 @@
 <Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">
 
     <PropertyGroup>
-        <TargetFramework>net6.0</TargetFramework>
+        <TargetFramework>net7.0</TargetFramework>
```
AndersenBell, line 4: the core.sample.shared project is multi-targeted, does this need to be too?

src/dymaptic.GeoBlazor.Core/Components/Layers/Graphic.cs
```diff
@@ -325,24 +337,64 @@ protected override void OnParametersSet()
...
+        if (JsModule is null) return;
 
-        await JsGraphicReference.InvokeVoidAsync("setAttributes", 
-            CancellationTokenSource.Token, Attributes);
+        await JsModule.InvokeVoidAsync("setGraphicAttributes", 
+            CancellationTokenSource.Token, Id, Attributes);
         ToSerializationRecord(true);
     }
+
+    private void GetJsModule()
```
AndersenBell, line 376: could you override the JSModule property and call this in the get body of it? then you don't need to explicitely call it. This also may not work with all of the serialization stuff going on

src/dymaptic.GeoBlazor.Core/Components/Layers/LayerObject.cs
```diff
@@ -36,10 +36,14 @@ public virtual async Task SetSymbol(Symbol symbol)
     {
         Symbol = symbol;
 
-        if (JsGraphicReference is not null)
+        if (JsModule is not null)
         {
-            await JsGraphicReference.InvokeVoidAsync("setSymbol", 
-                Symbol.ToSerializationRecord());
+            await JsModule.InvokeVoidAsync("setGraphicSymbol",
```
AndersenBell, line 41: is there any way to have a more concrete reference to these calls? Seems possible to mis-type them, or lose track of how many there are

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -1845,14 +1956,43 @@ protected override async Task OnParametersSetAsync()
...
         await UpdateView();
     }
+    
+    protected override async Task OnInitializedAsync()
+    {
+#if NET7_0_OR_GREATER
+        if (IsWebAssembly)
+        {
+#pragma warning disable CA1416
+            await JSHost.ImportAsync("arcGisJsInterop", "../_content/dymaptic.GeoBlazor.Core/js/arcGisJsInterop.js");
```
AndersenBell, line 1978: will this URL be fixed for every blazor project? SHould it be configurable?

TimPurdum: Yes, it's fixed.  I can't recall exactly how or why there is a `../` in the path, but `_content/projectNamespace` is the standard path for Blazor libraries, which takes you to the `wwwroot` in that library, and then you append `js` which is the folder.

src/dymaptic.GeoBlazor.Core/Components/Layers/LayerObject.cs
TimPurdum: 1. The called methods are in TypeScript, so I haven't discovered a way to verify them. We would probably have to write our own build-time method checker.
2. JsModule is normally a `CascadingParameter`, which means the one on `MapView` gets shared with all child components. However, since `Graphics` can regularly be added via code instead of razor components, I had to do a bit of juggling to make sure it was initialized in that class. For anything set in markup, it will cascade.

src/dymaptic.GeoBlazor.Core/Components/Layers/Graphic.cs
TimPurdum: I thought of that, I'll take another look. See my other answer for why this is required in `Graphic` but not elsewhere.

samples/dymaptic.GeoBlazor.Core.Sample.Wasm/dymaptic.GeoBlazor.Core.Sample.Wasm.csproj
TimPurdum: Since this is an .exe, multi-targeting isn't really necessary unless we want to run it in multiple runtimes. But that might be useful for testing, so I'll try it out.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/SerializationTests.razor
TimPurdum: It's a placeholder for the next step, where I need to replace all feature `queries` with protobuf serialization. I will add comments to it.

TimPurdum: Nope, sorry, you were right, this is old garbage. Deleting.

src/dymaptic.GeoBlazor.Core/Components/Geometries/SpatialReference.cs
```diff
@@ -225,5 +226,8 @@ public override void Write(Utf8JsonWriter writer, SpatialReference value, JsonSe
     }
 }
 
-internal record SpatialReferenceSerializationRecord([property:JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]int? Wkid)
+[ProtoContract(Name = "SpatialReference")]
```
morehavoc, line 229: It's most common in simple apps to have a SR only defined by a wkid, but the wkt is also important as that allows you to fully describe the SR. We may not need this rigth now, but let's at least make a gissue for it so we can support it.

AndersenBell (approved):

morehavoc (approved):

PR #160, opened 2023-04-13
Add View.ToMap and View.ToScreen conversion methods

### LINKED ISSUE #156: Add View.ToMap Method
MapView in ArcGIS has the ability to convert screen points to map cooridnates with `view.toMap(x, y)`. Add this method to the GeoBlazor MapView and test

TimPurdum:
Closes #156 
- Added both `ToMap` to convert a screen coordinates point to a map point and `ToScreen` to do the reverse
- Added a simple round trip test to the click even handler in the `Events.razor` samples page.


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Events.razor
```diff
@@ -147,9 +147,13 @@
 
 @code {
 
-    private void OnMapClick(ClickEvent clickEvent)
+    private async Task OnMapClick(ClickEvent clickEvent)
     {
         _clickJson = BuildViewableJson(clickEvent);
+        ScreenPoint screenPoint = await _view.ToScreen(clickEvent.MapPoint);
```
seahro, line 153: excited about putting this to use.  I will need to look some at the docs to see more if the same conversion methods will apply the same when drawing lines, polygons etc when multiple points are applied to the graphic, vs singular points

seahro (approved):
Looks good.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Events.razor
TimPurdum: Even though there is a `Point` referenced here, it is not necessarily a graphic, and the conversions would not be available for a line or polygon, but you could for example convert a single point _from_ a line/polygon.

PR #164, opened 2023-04-27
Feature/simple oauth

TimPurdum:
@morehavoc I have created a new PR so that you can review my changes to the OAuth flow.

I attempted to implement some of the ideas we discussed previously. What I discovered, however, is that `JSRuntime` is essentially never ready for calls until after a view has been initialized and rendered. Since we are using the JS `IdentityManager`, this means we simply can't "pre-load" authentication.

I added code in `MapView` that will trigger oauth initialization, and optionally trigger a login on first view render. However, you can also still call the methods you created in `OAuthAuthentication` directly if you are doing other work and don't want to show a map.

I created a Server OAuth sample to verify that the current pattern works for both Wasm and Server. The `AddScoped` method in DI works well for both, because there is only one effective scope in blazor WASM.

Unfortunately, I hit some roadblocks with a MAUI implementation. https://github.com/dotnet/maui/issues/2702 documents how OAuth is essentially broken for MAUI Windows/WinUI. Interestingly Morton from Esri commented on this with a workaround, probably what the ArcGIS .NET SDK team is using. I think it might make sense to have a MAUI-specific implementation in the future that uses the ArcGIS solution. But in the meantime, I'm comfortable with adding OAuth support _minus_ MAUI as long as we document it as such.


src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -2062,6 +2094,7 @@ protected override async Task OnInitializedAsync()
     protected override async Task OnAfterRenderAsync(bool firstRender)
     {
         ApiKey = Configuration["ArcGISApiKey"];
+        AppId = Configuration["ArcGISAppId"];
```
morehavoc, line 2097: Is this safe if either or both are not specified in the configuration?

src/dymaptic.GeoBlazor.Core/Model/OAuthAuthentication.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// <summary>
+    ///     Returns the current OAuth Token for use in requests.
+    /// </summary>
+    /// <returns>
+    ///     A string containing the current OAuth Token.
+    /// </returns>
+    public async Task<string> GetCurrentToken()
+    {
+        return await InvokeAsync<string>("getToken");
```
morehavoc, line 71: I was kind of hoping that we could centralize the "token" management such that the dev could just call "getCurrentToken" and get back the api key if that's the current token, or the oauth token if that's the current token. That way there is a central point to extract the current token if you need to pass it to a custom rest call.

docs/pages/authentication.md
```diff
@@ -40,11 +48,18 @@ your [ArcGIS Developer Dashboard](https://developers.arcgis.com/api-keys/).
...
 ## OAuth Sign-in
 
-Using a `ClientId` and `ClientSecret`, you can set up your application to use
-[OAuth Authentication](https://developers.arcgis.com/documentation/mapping-apis-and-services/security/oauth-2.0/).
-This secure system will redirect to an Esri login screen, and then back to your application, with a short-lived token
-that can be used in place of the API Token. This is useful for creating an application for a company or government
-agency that already has ArcGIS accounts.
+OAuth is a more secure way to authenticate your users with ArcGIS. It requires that your users have an ArcGIS account.
+You generate a new OAuth `Application` in the [ArcGIS Developer Dashboard](https://developers.arcgis.com/applications/),
+and then store the `ClientId` in your application configuration as `ArcGISAppId`.
```
morehavoc, line 53: Can we add a bit here about setting the Portal URL as well to support custom enterprise portal logins?

morehavoc (approved):
This looks pretty great tim! Just a couple questions/thoughts for you.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: Yes, the `IConfiguration` indexer is defined to return a nullable string if the value isn't found.

src/dymaptic.GeoBlazor.Core/Model/OAuthAuthentication.cs
TimPurdum: This was good feedback and pushed me to do a bit more refactoring. I added an `ApiKeyAuthentication` class, moved all the APIKey logic out of `MapView` and into this `LogicComponent`, and then created a new `AuthenticationManager` class that checks for both api keys and app ids, initializes the correct logic component, and exposes `GetCurrentToken` as you suggested.

TimPurdum:
@morehavoc I did another serious refactor, moving everything into `AuthenticationManager` to allow for handling both ApiKey and OAuth together, and support the idea of `GetCurrentToken`. If you have a chance to re-review this, please do.

src/dymaptic.GeoBlazor.Core/Model/AuthenticationManager.cs
```diff
@@ -0,0 +1,187 @@
...
+    /// <summary>
+    ///     The ArcGIS Enterprise Portal URL, only required if using Enterprise authentication.
+    /// </summary>
+    public string? PortalUrl
+    {
+        get
+        {
+            if (string.IsNullOrWhiteSpace(_portalUrl))
+            {
+                _portalUrl = _configuration["ArcGISPortalUrl"];
```
morehavoc, line 120: Do we have a default set for this somewhere? What I mean is I know we don't need it if it's just arcgis online, but maybe we should still have it set so that it's consistent?

morehavoc (approved):
This is great, I really like having this centralized this way!

src/dymaptic.GeoBlazor.Core/Model/AuthenticationManager.cs
TimPurdum: That might be consistent for every use of `AuthenticationManager`, but across ArcGIS JS and GeoBlazor, it is more common to leave optional parameters as `null`.

PR #167, opened 2023-05-07
Fix and document solution for wsv error

### LINKED ISSUE #144: Check on and Complete `Portal` Implementation
We have `Portal` (for setting the URL) implemented for basemaps, but not other things that can support an item. We want to support this to that a person could use their own portal and not just ArcGIS.com.  See this example from discord:
image

### LINKED ISSUE #166: Error on SceneViews 404 for stars.wsv
When loading a page with a `SceneView` in blazor server, we get the following error:

`https://samples.geoblazor.com/_content/dymaptic.GeoBlazor.Core/assets/esri/views/3d/environment/resources/stars.wsv 404 (Not Found)`

This is due to ArcGIS JS using an uncommon file type, `.wsv`, which is documented [here](https://developers.arcgis.com/javascript/latest/install-and-set-up/).

The solution includes adding support for wsv as a mime-type in `Program.cs`. Add to the server sample and documentation.

TimPurdum:
This removes the Blazor Server error about 404 for `stars.wsv` - Closes #166 

Also updated ArcGIS to 4.26.5, and updated the `assetCopy.ps1` file to auto-load the ArcGIS version from `package.json` so it needs only be updated in one place.

### DIFF
--- a/docs/pages/gettingStarted.md
+++ b/docs/pages/gettingStarted.md
@@ -76,9 +76,23 @@ nav_order: 2
    ```
 6. In `Program.cs`, add the following line to your `builder.Services` to inject logic components like `GeometryEngine`.
 
-```csharp
+   ```csharp
    builder.Services.AddGeoBlazor();
-```
+   ```
+
+   If you are using Blazor Server, you should also add the following lines to `Program.cs` to support the `.wsv` file type.
+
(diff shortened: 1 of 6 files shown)


AndersenBell (approved):
looks good

Issue #117, opened 2023-01-01
Add other methods to FeatureLayer


TimPurdum:
#115 and #116 added query and attachment methods to FeatureLayer, respectively.

Add the following other methods:

- applyEdits
- clone
- createPopupTemplate
- getFeatureType
- getField
- getFieldDomain
- refresh
- save
- saveAs

PR #169, opened 2023-05-12
Adding CSV Layer

AndersenBell:
Adding support for the CSV Layer

The sample pulls the sample CSV from the developer ArcGIS portal

It follows the GEOJson structure, so it sets the renderer and spatial reference in the interop js file

### DIFF
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Tests.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Tests.razor
@@ -8,6 +8,7 @@
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_graphicsLayer)">Graphics Layer</button>
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_geoJsonLayer)">GeoJSON Layer</button>
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_geoRssLayer)">GeoRSS Layer</button>
+<button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_csvLayer)">CSV Layer</button>
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_osmLayer)">OpenStreetMaps Layer</button>
 <h2>Markup Layers</h2>
 <button disabled="@(!_mapRendered)" @onclick="() => _showGraphicsMarkupLayer = !_showGraphicsMarkupLayer">Graphics Layer</button>
@@ -218,6 +219,8 @@
         new("https://services3.arcgis.com/GVgbJbqm8hXASVYi/ArcGIS/rest/services/World_Countries_(Generalized)/FeatureServer/0/query?where=1%3D1&outFields=*&f=geojson");
     private readonly GeoRSSLayer _geoRssLayer =
         new("https://disasterscharter.org/charter-portlets/cpi-mvc/activations/feed/rss/");
+    private readonly CSVLayer _csvLayer =
+        new("https://developers.arcgis.com/javascript/latest//sample-code/layers-csv/live/earthquakes.csv");
     private readonly OpenStreetMapLayer _osmLayer = new();
 
     private bool _mapRendered;
--- a/src/dymaptic.GeoBlazor.Core/Components/Layers/CSVLayer.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Layers/CSVLayer.cs
@@ -1,0 +1,65 @@
+﻿using Microsoft.AspNetCore.Components;
+using System.Text.Json.Serialization;
+
+
+namespace dymaptic.GeoBlazor.Core.Components.Layers;
+
+/// <summary>
+///     The CSVLayer is a point layer based on a CSV file (.csv, .txt). CSV is a plain-text file format used to
+///     represent tabular data, including geographic point features (latitude, longitude). Typically the latitude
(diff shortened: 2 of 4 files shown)


src/dymaptic.GeoBlazor.Core/Components/Layers/CSVLayer.cs
```diff
@@ -0,0 +1,65 @@
...
+    ///     referenced in a view, but its features will not be visible in the view.
+    /// </param>
+    /// <param name="listMode">
+    ///     Indicates how the layer should display in the LayerList widget. The possible values are listed below.
+    /// </param>
+    public CSVLayer(string url, string? title = null,
+        double? opacity = null, bool? visible = null, ListMode? listMode = null)
+    {
+        Url = url;
+        Title = title;
```
TimPurdum, line 49: I realize this is missing in a _lot_ of classes, including most layers, but we should wrap these parameter setters in `#pragma warning disable BL0005`. Search for that and you'll see some examples where I've implemented it. If you don't do this, and you create the class in C# with this constructor, you will get a warning about setting parameters outside of markup.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Tests.razor
```diff
@@ -8,6 +8,7 @@
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_graphicsLayer)">Graphics Layer</button>
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_geoJsonLayer)">GeoJSON Layer</button>
 <button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_geoRssLayer)">GeoRSS Layer</button>
+<button disabled="@(!_mapRendered)" @onclick="() => ToggleLayer(_csvLayer)">CSV Layer</button>
```
TimPurdum, line 11: Can you create a full samples page for this new layer type? It helps with the discoverability, and this tests page is a mess that I don't publish to our `samples.geoblazor.com` site.

TimPurdum (changes requested):
Thanks for doing this! I think we need some new gissues to investigate properties and methods that are not implemented here, and on some other common layer types like `GeoJSON`, but are in the ArcGIS spec. If you have time to write up some of those, feel free. An example of how I've done that is https://github.com/dymaptic/GeoBlazor/issues/117.

Request changes mostly just for the dedicated sample page.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/CSVLayers.razor
```diff
@@ -0,0 +1,53 @@
...
+@code {
+
+    private void OnMapRendered()
+    {
+        _mapRendered = true;
+    }
+
+    private async Task AddLayer()
+    {
+        if (!string.IsNullOrEmpty(_csvLayerUrl))
```
TimPurdum, line 36: Should we remove the original CSV layer when adding a new one? Did you try binding to the `<CSVLayer Url="_csvLayerUrl">` value? I'm sure there could be issues with that, you might have to call `mapView.Refresh()`.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/razorCopy.ps1
```diff
@@ -1,2 +1,7 @@
-﻿Get-ChildItem -Path .\Pages -Filter "*.razor" -File |
+﻿$path = ".\wwwroot\pages\"
+If(!(test-path -PathType container $path))
+{
+      New-Item -ItemType Directory -Path $path
+}
```
TimPurdum, line 5: Thanks for adding this!

src/dymaptic.GeoBlazor.Core/Components/Layers/CSVLayer.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace dymaptic.GeoBlazor.Core.Components.Layers;
+
+/// <summary>
+///     The CSVLayer is a point layer based on a CSV file (.csv, .txt). CSV is a plain-text file format used to
+///     represent tabular data, including geographic point features (latitude, longitude). Typically the latitude
+///     coordinate is the Y value, and the longitude coordinate is the X value. The X, Y coordinates must be stored
+///     in SpatialReference.WGS84 in csv feed.
+///     <a target="_blank" href="https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-CSVLayer.html">
+///         ArcGIS
```
TimPurdum, line 13: These line breaks cause issues with our docs generation, I believe. We might need to inline this.

TimPurdum (approved):
Looks great!

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/CSVLayers.razor
AndersenBell: I didn't try it. I was trying to include multiple examples of adding a csv layer, so if you look at the source you can see how to do it declaratively and programmatically.

src/dymaptic.GeoBlazor.Core/Components/Layers/CSVLayer.cs
AndersenBell: ok, I think it was auto-formatted this way

AndersenBell: It looks like its formatted exactly like other layers in the project

PR #171, opened 2023-05-13
Feature/163 geometry methods

### LINKED ISSUE #163: Add missing methods to Extent
The GeoBlazor Extent class does not support all of the methods available to the [esri/geometry/Extent](https://developers.arcgis.com/javascript/latest/api-reference/esri-geometry-Extent.html) class in the ArcGIS JavaScript API.  Methods like [clone()](velopers.arcgis.com/javascript/latest/api-reference/esri-geometry-Extent.html#clone) and [expand()](https://developers.arcgis.com/javascript/latest/api-reference/esri-geometry-Extent.html#expand) are missing.

## Use case ##
I am drawing bus routes on a map. I create an extent based on the locations of bus stops. I want the Extent to cover a slightly larger area than the calculated Extent.

I should be able to syntax like the following:

```csharp
mapview.GoTo(SomeExtent.Clone().Expand(1.2);
```

## Work Around ##
Create a new Extent in C# code and calculate the Extent lat/lng from the source Extent.

Tested under Windows 11 with Windows and Android as target platforms. Compiled with VS 2022 17.5.3 and GeoBlazor 2.1.0

TimPurdum:
Closes #163 

- Implements all missing geometry-related methods, but moves them all to the `GeometryEngine`
- Documents implementation and how it differs from ArcGIS in new docs page
- New samples page `GeometryMethods.razor` tests many of the new methods in `GeometryEngine`


AndersenBell (approved):
looks good

PR #177, opened 2023-05-20
Fix widgets on refresh

### LINKED ISSUE #176: Widgets do not re-render on `MapView.Refresh()`
Reported in Discord. When a call is placed to `MapView.Refresh()`, any existing widgets are not re-rendered. This is due to not clearing out the `arcGisObjectRefs` dictionary on re-rendering

TimPurdum:
Closes #176 

Simply removed the check in the `arcGisObjectRefs` dictionary that was preventing re-creating widgets. I checked and none of the calls to `createWidget` seem susceptible to rapid-calling from the blazor render cycle, so this should be fine. We can't re-use the stored widgets, however, in case the properties of the widget were changed and the goal of the refresh is to update it.

On testing on `Widgets.razor` in the samples app, I noticed that I also needed to add logic to clear out existing widgets placed outside of the map view in container divs.

### DIFF
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Widgets.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Widgets.razor
@@ -40,6 +40,7 @@
         <div id="gallery-box"></div>
     </div>
 </div>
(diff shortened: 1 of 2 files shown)


seahro (approved):
straight forward

PR #180, opened 2023-05-29
Feature/179 pro beta

### LINKED ISSUE #179: Update to support Pro beta
We are preparing to release a beta version of the GeoBlazor Pro library. Update all necessary dependencies in Core to support the current Pro version.

TimPurdum:
Closes #179 

- Adds support for Pro library beta release
- Minor code cleanups
- Tested all samples


src/dymaptic.GeoBlazor.Core/Scripts/featureLayer.ts
```diff
@@ -75,14 +110,42 @@ export default class FeatureLayerWrapper {
         return await this.layer.queryObjectIds(jsQuery, options);
     }
 
-    async queryRelatedFeatures(query: DotNetRelationshipQuery, options: any, dotNetRef: any): Promise<FeatureSet | null> {
+    async queryRelatedFeatures(query: DotNetRelationshipQuery, options: any, dotNetRef: any, viewId: string | null)
```
seahro, line 113: so the `queryFeatures` and the `queryRelatedFeatures` methods do essentially the same thing since they are almost identical with the primary difference being the use of the `graphicsDictionary` and `featureSetsDictionsary` in `queryRelatedFeatures` because we are following the featureSets relationships vs the features themselves?

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Scene.razor
```diff
@@ -132,5 +133,6 @@
     private SceneView? _view;
     private bool _rendered;
 
+    [SuppressMessage("ReSharper", "NotAccessedPositionalProperty.Local")]
```
seahro, line 136: I had to look this up.  Is there a reason to keep the testObject in this file vs put in the Shared/Pages/Tests.razor or some other file?

seahro (approved):
Couple questions.  Overall, no specific issues.

src/dymaptic.GeoBlazor.Core/Scripts/featureLayer.ts
TimPurdum: I don't have a ton of knowledge around this, I would defer to Esri's documentation or others on what the different queries mean.

PR #183, opened 2023-06-15
Fix popup handlers and removing graphics

### LINKED ISSUE #181: Swallowed exception and failure on MapView.RemoveGraphic
When trying to remove a graphic from the View directly via `MapView.RemoveGraphic`, there was an exception that was being caught and written to the console, but the actual removal also fails. 

Discovered the cause was mismatched parameters sent to the JavaScript method.

### LINKED ISSUE #182: Popup Action Handlers added repeatedly
When a `PopupWidget` or `PopupTemplate` has an `Action` associated with it, the JavaScript handler is added every time the popup renders, without removing the old ones. This can cause issues with the associated callback method firing repeatedly.

TimPurdum:
Closes #181 
Closes #182 

Allows user to remove a graphic successfully from the `MapView`
Prevents multiple registrations of the event handler in a `PopupWidget` or `PopupTemplate`


src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
```diff
@@ -227,7 +227,13 @@ export function buildJsPopupTemplate(popupTemplateObject: DotNetPopupTemplate, v
     if (viewId !== null) {
         let view = arcGisObjectRefs[viewId] as View;
         if (hasValue(view)) {
-            view.popup.on("trigger-action", async (event: PopupTriggerActionEvent) => {
+            if (hasValue(triggerActionHandler)) {
```
seahro, line 230: do we want to start to convert more null checks over to 'hasValue'? I see you have included this multiple times here and above, but still have the null checks?

seahro (approved):
Looks good one question about the hasValue vs null checks.

src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
TimPurdum: It would be a good refactoring to do, to move to always using `hasValue`.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -1533,21 +1533,14 @@ public async Task SolveServiceArea(string serviceAreaUrl, double[] driveTimeCutO
     /// </param>
     public async Task RemoveGraphic(Graphic graphic)
     {
-        try
```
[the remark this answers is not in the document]
TimPurdum: This try/catch was swallowing the exception and not exposing it to the end user, even though there was a real problem. I sometimes intentionally swallow exceptions when they are caused by say a user navigating away from the map, but in this case it was misguided.

PR #187, opened 2023-07-06
fix: add pattern matching for ios support

### LINKED ISSUE #186: iOS Release Builds: Unable to `GetActiveEventHandlers` due to JIT limitation
## Steps to Reproduce
Build MAUI Blazor App w/ Geoblazor Map
Deploy for Release to iOS (locally or Testflight)
Load Map

## What Happens
An exception is thrown
```
System.ExecutionEngineException: Attempting to JIT compile method '(wrapper dynamic-method) object object:Thunk1ret_Object_CallSite_Boolean_Object (System.Func`2<object[], object>,System.Runtime.CompilerServices.CallSite,bool,object)' while running in aot-only mode. See https://docs.microsoft.com/xamarin/ios/internals/limitations for more information.
...
 at dymaptic.GeoBlazor.Core.Components.Views.MapView.GetActiveEventHandlers()
 at dymaptic.GeoBlazor.Core.Components.Views.MapView.<RenderView>b__308_2()
 at Microsoft.Maui.Dispatching.DispatcherExtensions.<>c__DisplayClass3_0.<<DispatchAsync>b__0>d.MoveNext()
 ```
 
In the `GetActiveEventHandlers` method, a `dynamic` object is used which is not allowed on iOS with .NET (noted here - https://learn.microsoft.com/en-us/xamarin/ios/internals/limitations#systemreflectionemit).  This was the only instance of `dynamic` I could find in the project.

I have PR incoming shortly to fix this with pattern matching instead, but I wanted to open the issue for tracking.

/cc @anotherlab

### PR DESCRIPTION
**Pull request by nwestfall**
Fixes #186 

Instead of using `dynamic` in `GetActiveEventHandlers`, we can detect if it's a callback using pattern matching.  This is allowed in iOS.

https://learn.microsoft.com/en-us/xamarin/ios/internals/limitations#systemreflectionemit

I tested using the Server and WASM projects, all events still show.
image

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
@@ -2257,15 +2257,13 @@ public partial class MapView : MapComponent
     {
         List<string> activeHandlers = new();
 
-        IEnumerable<PropertyInfo> funcCallbacks = GetType()
-            .GetProperties()
-            .Where(p => p.PropertyType.Name.StartsWith("Func"));
+        var properties = GetType().GetProperties();
+
+        IEnumerable<PropertyInfo> funcCallbacks = properties.Where(p => p.PropertyType.Name.StartsWith("Func"));
 
         activeHandlers.AddRange(funcCallbacks.Select(x => x.Name));
 
-        IEnumerable<PropertyInfo> eventCallbacks = GetType()
-            .GetProperties()
-            .Where(p => p.PropertyType.Name.StartsWith(nameof(EventCallback)));
+        IEnumerable<PropertyInfo> eventCallbacks = properties.Where(p => p.PropertyType.Name.StartsWith(nameof(EventCallback)));
 
         foreach (PropertyInfo callbackInfo in eventCallbacks)
         {
--- a/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
@@ -2264,19 +2264,31 @@ public partial class MapView : MapComponent
 
         foreach (PropertyInfo callbackInfo in callbacks)
         {
-            dynamic? callback = callbackInfo.GetValue(this);
+            object? callback = callbackInfo.GetValue(this);
 
-            try
+            if(callback != null)
             {
-                if (callback is not null && callback.HasDelegate)
+                var hasDelegate = callback switch
                 {
-                    activeHandlers.Add(callbackInfo.Name);
-                }
-            }
-            catch
-            {
-                // Funcs don't have "HasDelegate"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs [hunk 1]
```diff
@@ -2264,19 +2264,31 @@ protected List<string> GetActiveEventHandlers()
 
         foreach (PropertyInfo callbackInfo in callbacks)
         {
-            dynamic? callback = callbackInfo.GetValue(this);
+            object? callback = callbackInfo.GetValue(this);
```
TimPurdum, line 2267: @nwestfall this is great! I have a few suggestions on how we can make it more streamlined.

- On line 2260, we could actually pull the `EventCallback`s and `Func`s separately as propertyInfos, instead of together. This would eliminate the need to check if it's a Func later.
- There is an interface, `IEventCallback` that all the `EventCallback`s and `EventCallback<T>`s inherit from. I think we could just cast to that instead of each return type. This would make it more extensible, as with the current version, if we add new EventCallbacks with different types, we have to remember to update this method.

I think I'd like to see a refactor using one or both of these ideas before we merge. Let me know if you would like me to implement that myself, or you are welcome to update this branch and PR.

Thanks again for finding this and working on it!

TimPurdum (changes requested):

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
nwestfall: @TimPurdum  I'll look at your first point.  For the second point, I actually tried this but it's an internal interface.

https://github.com/dotnet/aspnetcore/blob/1fbe264736d62fab1c37f0986df2e90fd4ab33b3/src/Components/Components/src/IEventCallback.cs#L7

TimPurdum: Ah, I missed that. Bummer!

TimPurdum: If we can't find a way around enumerating the callback `T`s, I'm ok with keeping it.

nwestfall: The only way I could think of doing it (and maybe this is longer term), is if all the GeoBlazor Events (`ClickEvent`, `BlurEvent`), implemented an interface.  Would have to test, but maybe then you could do `EventCallback<IGeoBlazorEvent>` to catch all of them.

nwestfall: Doesn't look like you can :(, so I don't have any other ideas at this time.

TimPurdum (approved):

PR #189, opened 2023-07-07
prevent esbuild when building from Pro

TimPurdum:
Refactor to support Pro package, prevent double-build/bundle of JS

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
+++ b/src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
@@ -69,36 +69,37 @@
 
     <ItemGroup>
         <Folder Include="wwwroot\assets" />
+        <Content Remove="node_modules\**" />
     </ItemGroup>
 
-    <Target Name="NPM Install" BeforeTargets="PreBuildEvent" Condition="$(TargetFrameworks.StartsWith($(TargetFramework)))">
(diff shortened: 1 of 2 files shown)


src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
```diff
@@ -69,36 +69,37 @@
 
     <ItemGroup>
         <Folder Include="wwwroot\assets" />
+        <Content Remove="node_modules\**" />
     </ItemGroup>
 
-    <Target Name="NPM Install" BeforeTargets="PreBuildEvent" Condition="$(TargetFrameworks.StartsWith($(TargetFramework)))">
+    <Target Name="NPM Install" BeforeTargets="PreBuildEvent" Condition="$(TargetFrameworks.StartsWith($(TargetFramework))) AND '$(OptOutFromCoreEsBuild)' != 'true'">
```
AndersenBell, line 75: does OptOutFromCoreEsBuild need to be set anywhere in this project? or will it default to true?

AndersenBell (approved):
looks good

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
TimPurdum: It defaults to `false`, which we want here because normally you want all this to run. It is only `true` when running from the Pro repository, where this will be set to `true` to skip these steps.

PR #196, opened 2023-07-13
Adding measurement widget

### LINKED ISSUE #58: Add Measurement Widget
[Measurement](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Measurement.html)

1. Add basic features to implement the sample as in JS docs.
2. Add sample to existing Widgets page or new sample page.

AndersenBell:
closes #58 

Adds measurement widget to the map

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
@@ -37,6 +37,14 @@ public abstract class Widget : MapComponent
     public string? ContainerId { get; set; }
 
     /// <summary>
+    /// Icon which represents the widget. It is typically used when the widget is controlled by another one (e.g. in the Expand widget).
+    /// <a href="https://developers.arcgis.com/calcite-design-system/icons/">Calcite Icons</a>
+    /// </summary>
+    [Parameter]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Icon { get; set; }
+
+    /// <summary>
     ///     The type of widget
     /// </summary>
     [JsonPropertyName("type")]
--- a/src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
@@ -37,14 +37,6 @@ public abstract class Widget : MapComponent
     public string? ContainerId { get; set; }
 
     /// <summary>
-    /// Icon which represents the widget. It is typically used when the widget is controlled by another one (e.g. in the Expand widget).
-    /// <a href="https://developers.arcgis.com/calcite-design-system/icons/">Calcite Icons</a>
-    /// </summary>
-    [Parameter]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string? Icon { get; set; }
-
-    /// <summary>
     ///     The type of widget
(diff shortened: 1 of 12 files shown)


src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
```diff
@@ -36,6 +36,14 @@ public abstract class Widget : MapComponent
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string? ContainerId { get; set; }
 
+    /// <summary>
+    /// Icon which represents the widget. It is typically used when the widget is controlled by another one (e.g. in the Expand widget).
+    /// <a href="https://developers.arcgis.com/calcite-design-system/icons/">Calcite Icons</a>
+    /// </summary>
+    [Parameter]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? Icon { get; set; }
```
seahro, line 45: Does this need to be in the `MeasurementWidget.cs`? I saw that the both have Icon properties.  Just wanted to verify.

seahro (approved):
Looks good.  One question on the Icon property that is in widget.

src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
AndersenBell: it should be removed from the measurement widget and inherited from the widget class

src/dymaptic.GeoBlazor.Core/Objects/ArealUnit.cs
```diff
@@ -22,6 +23,58 @@ public enum ArealUnit
...
+
+/// <summary>
+/// Units for area measurement. Use one of the possible values listed below or any of the numeric codes for area units.
+///  <a target="_blank" href="https://developers.arcgis.com/javascript/latest/api-reference/esri-core-units.htm">
+///         ArcGIS
+///         JS API
+///     </a>
+/// </summary>
+[JsonConverter(typeof(EnumToKebabCaseStringConverter<AreaUnit>))]
+public enum AreaUnit
```
TimPurdum, line 39: We need a better way to identify the units used in GeometryEngine vs. the ones used in the new widget. Since they changed the names on me anyways, could we create a new `GeometryEngineAreaUnit`, create overloads in `GeometryEngine.cs` that take those instead, and mark the ones that take `ArealUnit` as `[Obsolete]`?

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -1755,6 +1756,16 @@ async function createWidget(widget: any, viewId: string): Promise<Widget | null>
...
             break;
+        case 'measurement':
+            newWidget = new Measurement({
+                view: view,
+                activeTool: widget.activeTool,
+                areaUnit: widget.areaUnit,
+                linearUnit: widget.linearUnit,
+                label: widget.label,
+                icon: widget.icon,
+            });
```
TimPurdum, line 1767: I think this will break if you don't set all the values in the sample. The pattern I suggest is `activeTool: widget.activeTool ?? undefined`. This prevents setting `null` values in arcgis, which are sometimes not expected or handled.

TimPurdum (changes requested):
Looking good, just a few requests!

TimPurdum (approved):
Nicely done!

PR #198, opened 2023-07-14
Extra fixes for Pro widgets

TimPurdum:
A few extra things from the Pro work
- Added `Icon` and `WidgetId` (aka JS Id) to `Widget`
- Fixed the reason that the arcgis assets were not updating for me in Core.csproj
- Added a debug catch around view.popup.on not being a thing when using the sketch tool (needs further investigation later).


seahro (approved):
Looks good.  I have seen this `popup.on` error only a couple of times (it seem infrequent).

PR #208, opened 2023-08-02
fix popup actions

### LINKED ISSUE #207: Popup Actions not working
Reported in discord, the popup actions are not working. In the dev tools, we see the error: `view.popup.on is not a function`.

TimPurdum:
Closes #207 
The `view.popup.on` function is not available until the view is rendered. Rather than do some sort of re-registration of the `popupTemplate`, I realized we could use a `reactiveUtils.once` to register the handler after the popup is ready.

Also in this PR
- A few `undefined`s added to a `jsBuilder` to prevent null errors
- Had to update the `copyAssets.ps1`. Powershell's default behavior seems to have changed. Christopher and I saw this before, now my version matches what his Mac was doing 3 months ago...
- removed an old project reference from the solution file

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
+++ b/src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
@@ -110,6 +110,7 @@ import ComboBoxInput from "@arcgis/core/form/elements/inputs/ComboBoxInput";
 import RadioButtonsInput from "@arcgis/core/form/elements/inputs/RadioButtonsInput";
 import SwitchInput from "@arcgis/core/form/elements/inputs/SwitchInput";
 import Domain from "@arcgis/core/layers/support/Domain";
+import * as reactiveUtils from "@arcgis/core/core/reactiveUtils";
 
 
 export function buildJsSpatialReference(dotNetSpatialReference: DotNetSpatialReference): SpatialReference {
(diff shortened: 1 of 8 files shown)


src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
```diff
@@ -727,9 +731,9 @@ export function buildJsQuery(dotNetQuery: DotNetQuery): Query {
         where: dotNetQuery.where ?? "1=1",
         spatialRelationship: dotNetQuery.spatialRelationship as any ?? "intersects",
         distance: dotNetQuery.distance ?? undefined,
-        units: dotNetQuery.units as any ?? null,
```
AndersenBell, line 730: https://developers.arcgis.com/javascript/latest/api-reference/esri-rest-support-Query.html#units

The default value is null per the API?

src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
```diff
@@ -727,9 +731,9 @@ export function buildJsQuery(dotNetQuery: DotNetQuery): Query {
...
+        units: dotNetQuery.units as any ?? undefined,
         returnGeometry: dotNetQuery.returnGeometry ?? false,
-        outFields: dotNetQuery.outFields ?? null,
```
AndersenBell, line 732: Same here, default value is null:
https://developers.arcgis.com/javascript/latest/api-reference/esri-rest-support-Query.html#outFields

AndersenBell (approved):
Just a couple comments on the default values that got updated.

seahro (commented):
Looks good.  I had one question about possibly adding a note in the `reactiveUtils` or `popup` documentation about this method

seahro (approved):

agnesstelmach (approved):
TypeScript looks good to me :)

src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
TimPurdum: Yes, I haven't actually improved anything here, but I also didn't hurt anything. I think I made this change as part of an investigation into an issue, and it seemed to make more sense to me to keep it `undefined` as more consistent.

TimPurdum:
> Looks good. I had one question about possibly adding a note in the reactiveUtils or popup documentation about this method
@seahro I don't think your actual question came through. However, I've been going off the thought that only the C# layer needs documentation, but that was for the public-facing customer. We can certainly add comments for the internal maintainers. I'll see about adding one here.

PR #216, opened 2023-08-03
Updating feature layer Symbol rendering

### LINKED ISSUE #215: Feature Layer not displaying when Symbol is specified and Extent updated
Found on the samples application.
When a FeatureLayer is added to a view and then the extent is modified, if a new feature layer is added (or the current one is modified or removed/added) it will no longer render because the Symbol object is not populated after extent update

AndersenBell:
closes #215 

Allows the feature layer to apply the Symbol when it has not been sent.


TimPurdum (approved):
Thanks for investigating this!

PR #217, opened 2023-08-03
Bug/214 sql filter

### LINKED ISSUE #214: SQL Filter Sample Failing
https://samples.geoblazor.com/sql-filter-query

TimPurdum:
Closes #214 , fixing the sql filter page, by reverting some bad changes to `arcGisJsInterop.updateLayer`. Basically, FL's from the server (url or portal Id) were not getting some updates. I _think_ the original change was to avoid an issue with trying to jam a `SpatialReference` on a server feature layer, so I left that property client-side only for updates. I also tested with core and pro samples, and could not recreate the original problem after this reversion.

Also
- Fixed a Pro bug caused by the change in #208. We need to handle both the `view.popup.on` existing and not existing in `jsBuilder.buildJsPopupTemplate`
- While testing in Pro, on `ApplyEdits`, it tries to pass back the FeatureLayer, but it is apparently huge, even without graphics loaded. I noticed that `Fields` were being passed as-is from ArcGIS, and the `Renderer` I don't think needed to be passed.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Widgets/Widget.cs
@@ -68,6 +68,13 @@ public abstract class Widget : MapComponent
     {
         JsObjectReference = jsObjectReference;
     }
+    
+    /// <inheritdoc />
+    protected override void OnParametersSet()
+    {
+        base.OnParametersSet();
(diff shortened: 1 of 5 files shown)


AndersenBell (commented):
looks good

AndersenBell (approved):

PR #223, opened 2023-08-17
Bug/222 popup action update

### LINKED ISSUE #222: PopupAction Investigation
A user on discord reported the inability to set a popup action "on the fly" using `graphic.SetPopupTemplate`. This gissue is to track progress on that. Another user also requested a sample of just adding a point to the map on click, so I will add that to test the method.

TimPurdum:
Closes #222 

- Tested against the new sample on this branch and the samples that were shared on Discord
- Added the `id` to the `PopupTemplateSerializationRecord` for protobuf
- Added the same field to `graphic.json` so the JS side has it for deserialization
- Implemented a `lookup` function in `jsbuilder.ts` and `MapView.razor.cs` that finds a missing dotnet object reference
- Also added missing constructors for `ActionButton` and `ActionToggle`

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
@@ -969,6 +969,49 @@ public partial class MapView : MapComponent
     /// </summary>
     [Parameter]
     public int? GraphicSerializationChunkSize { get; set; }
+    
+    /// <summary>
+    ///     For internal use only, this looks up a missing <see cref="DotNetObjectReference"/> for a <see cref="PopupTemplate"/>
+    ///     and returns it to JavaScript.
+    /// </summary>
+    [JSInvokable]
+    public DotNetObjectReference<PopupTemplate>? GetDotNetPopupTemplateObjectReference(Guid popupTemplateId)
+    {
+        foreach (Graphic graphic in Graphics)
+        {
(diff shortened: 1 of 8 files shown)


src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -969,6 +969,49 @@ public async Task OnJavascriptLayerViewCreateError(LayerViewCreateErrorEvent err
...
+                foreach (Graphic graphic in featureLayer.Source)
+                {
+                    if (graphic.PopupTemplate?.Id == popupTemplateId)
+                    {
+                        return graphic.PopupTemplate.DotNetPopupTemplateReference;
+                    }
+                }
+            }
+
+            if (layer is GraphicsLayer graphicsLayer)
```
AndersenBell, line 1001: could this be an else if?

AndersenBell (approved):
looks good

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: It could. I was thinking it doesn't change the flow at all (I use this pattern when `return` is in each `if`), but in this case, you could find a `FeatureLayer`, not find the popup, and then have to ask again what it's layer type is, so you're right.

PR #224, opened 2023-08-22
Feature/66 add imagery layer

### LINKED ISSUE #66: Add an ImageryLayer
[ImageryLayer](https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-ImageryLayer.html)

1. Add the `ImageryLayer`
2. Add the layer to `LayerTypes.razor` created in #63. Alternatively, if these are too different, create a new samples page.

seahro:
Partially Closes #66 

This is the core portion of the imagery layer gissue #66 .  
In this portion of the imageryLayer is primarily: the imagery layer sample, `imageryLayer.cs`, and the minimal implementation of the imagerylayer in `arcGisJsInterop.ts`.  
The other raster classes (RasterShadedReliefRenderer, RasterColormapRenderer, VectorFieldRenderer, and FlowRenderer) and the related methods in `arcGisJsInterop` will be activated and reside in the Pro repositiory.  Finalizing gissue#66 will implement a more comprehensive sample page that uses the different raster renderers in the Pro to display the different imagery layer rendering capability.


src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
```diff
@@ -0,0 +1,143 @@
...
+    {
+    }
+
+    /// <summary>
+    ///     Constructor for use in code
+    /// </summary>
+    /// <param name="url"></param>  
+    /// <param name="portalItem"></param>
+    /// <param name="renderer"></param>
+    /// <param name="format"></param>
```
TimPurdum, line 37: Either fill out or delete the `<param>` lines please

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            case PortalItem portalItem:
+                if (!portalItem.Equals(PortalItem))
+                {
+                    PortalItem = portalItem;
+                    LayerChanged = true;
+                }
+
+                break;
+            //case IImageryRenderer renderer:
```
TimPurdum, line 84: What happened with the renderer? We're not supporting it? I think you were working on implementing it in GB.Pro, but this code would still need to be uncommented.

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
```diff
@@ -0,0 +1,143 @@
...
+        PortalItem?.ValidateRequiredChildren();
+        //Renderer?.ValidateRequiredChildren();
+    }
+}
+
+/// <summary>
+/// The format of the data sent by the server.
+/// </summary>
+[JsonConverter(typeof(EnumToKebabCaseStringConverter<Format>))]
+public enum Format
```
TimPurdum, line 129: This enum has a few issues
- Doesn't appear to be used in this layer, so if it is used at all, it should be co-located with the using layer/component
- The name is probably too vague and will collide with other types in the future. I would suggest naming it something like `ImageFormat` for clarity. You can still name a property as `Format` with type `ImageFormat` for serialization purposes.

src/dymaptic.GeoBlazor.Core/Components/Renderers/IImageryRenderer.cs
```diff
@@ -0,0 +1,19 @@
+﻿using dymaptic.GeoBlazor.Core.Components.Layers;
+
+namespace dymaptic.GeoBlazor.Core.Components.Renderers;
+
+/// <summary>
+///     This interface is to bridge . 
+///     <a target="_blank" href="https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-ImageryLayer.html">
```
TimPurdum, line 7: finish comment on line 6, update (or remove) link so it points to the renderer docs

src/dymaptic.GeoBlazor.Core/Components/Renderers/IImageryRenderer.cs
```diff
@@ -0,0 +1,19 @@
...
+///         JS API
+///     </a>
+/// </summary>
+public interface IImageryRenderer
+{
+
+    /// <summary>
+    ///     The method to enable the Renderer to validate.
+    /// </summary>
+    void ValidateRequiredChildren();
```
TimPurdum, line 18: This probably points to why you commented out the `RegisterChildComponent` code. I think the solution is to make this an abstract base class that inherits from `Renderer` or `MapComponent`, called `ImageryRenderer`, and then it will have all the expected properties and methods.

src/dymaptic.GeoBlazor.Core/Components/Renderers/RasterStretchRenderer.cs
```diff
@@ -155,6 +155,10 @@ internal override void ValidateRequiredChildren()
         base.ValidateRequiredChildren();
     }
 
+    void IImageryRenderer.ValidateRequiredChildren()
```
TimPurdum, line 158: replace the interface with a base class (see notes there) and then you shouldn't need this duplicate `ValidateRequiredChildren`

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -2183,6 +2186,25 @@ export async function createLayer(layerObject: any, wrap?: boolean | null, viewI
...
+            //    imageryLayer.renderer = buildJsRasterStretchRenderer(layerObject.renderer) as RasterStretchRenderer;
+            //}
+            copyValuesIfExists(layerObject, 'bandIds', 'blendMode', 'capabilities', 'compressionQuality', 'compressionTolerance',
+                'copyright', 'customParameters', 'definitionExpression', 'effect', 'fields', 'fieldsIndex', 'format',
+                'hasMultidimensions', 'imageMaxHeight', 'imageMaxWidth', 'interpolation', 'legendEnabled', 'maxScale', 'minScale',
+                'mosaicRule', 'multidimensionalInfo', 'multidimensionsionalSubset', 'noData', 'noDataInterpretation', 'objectIdField',
+                'pixelFilter', 'pixelType', 'popupEnabled', 'popupTemplate', 'rasterFields', 'rasterFunction', 'rasterFunctionInfos',
+                'refreshInterval', 'renderingRule', 'serviceRasterInfo', 'sourceJSON', 'spatialReference', 'timeExtent', 'timeInfo',
+                'timeOffset', 'useViewTime', 'version')
+
```
TimPurdum, line 2205: This giant list is suspicious. There are a lot of complex types that need pulled out into builders, such as `popupTemplate`, `timeExtent`, `spatialREference`, etc. Only simple types (string, number, boolean) are going to work correctly in a direct copy.

src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
```diff
@@ -552,6 +558,116 @@ export function buildJsRenderer(dotNetRenderer: any): Renderer | null {
     return dotNetRenderer
 }
 
+//export function buildJsRasterColormapRenderer(dotNetRasterColormapRenderer: DotNetRasterColormapRenderer): RasterColormapRenderer | null {
```
TimPurdum, line 561: why is this all commented out?

TimPurdum (changes requested):
Good start!

src/dymaptic.GeoBlazor.Core/Components/Renderers/RasterStretchRenderer.cs
seahro: removed and successfully tested

src/dymaptic.GeoBlazor.Core/Components/Renderers/IImageryRenderer.cs
seahro: corrected this class to switch to an abstract class and corrected the `registerchildcomponent`, `unregisterchildcomponent`, and validatechildcomponent`override methods

src/dymaptic.GeoBlazor.Core/Components/Renderers/IImageryRenderer.cs
seahro: corrected the href to point at the renderer section

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
seahro: adjusted

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
seahro: turned the renderer back on.  I think I had shut it off while troubleshooting an item.

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
seahro: filled out the comments and removed the format parameter that was removed.

src/dymaptic.GeoBlazor.Core/Scripts/jsBuilder.ts
seahro: This code was commented out while working on the one imagery layer renderer, prior to testing to the other renderers.  I believe this was an oversight in un commenting this code prior to pushing the code up.

Code is now uncommented and successfully tested.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
seahro: The imagery renderers have various properties that they reference in the imagerylayer.  I think this can be pared down further after testing with the other 5 renderers in GB Pro.

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
```diff
@@ -0,0 +1,148 @@
...
+    ///     The <see cref="PortalItem" /> from which the layer is loaded.
+    /// </param>
+    /// <param name="renderer">
+    ///     An abstract base class that implements a variety of imagery renderers.
+    /// </param>
+
+    public ImageryLayer(string? url = null, PortalItem? portalItem = null, ImageryRenderer? renderer = null
+        ) 
+    {
+        if (url is null && portalItem is null)
```
TimPurdum, line 47: We need the `#pragma warning disable BL0005`, such as you can see on `Point.cs` to silence setting Blazor `[Parameters]` in code in these constructors. It looks like it is missing from all the Layer types, can you add that?

src/dymaptic.GeoBlazor.Core/Components/Renderers/ImageryRenderer.cs
```diff
@@ -0,0 +1,19 @@
...
+///         JS API
+///     </a>
+/// </summary>
+public abstract class ImageryRenderer: MapComponent
+{
+
+    /// <summary>
+    ///     The method to enable the Renderer to validate.
+    /// </summary>
+    //void ValidateRequiredChildren();
```
TimPurdum, line 18: remove unneeded code, and add a comment in the XML at the top to clarify that this is an abstract class.

src/dymaptic.GeoBlazor.Core/Components/Renderers/RasterStretchRenderer.cs
```diff
@@ -155,6 +155,10 @@ internal override void ValidateRequiredChildren()
         base.ValidateRequiredChildren();
     }
 
+    //void ImageryRenderer.ValidateRequiredChildren()
```
TimPurdum, line 158: remove

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: My point in the first comment here was that properties with complex types should never be added to `copyValuesIfExists` in the first place, because it will silently fail. I wasn't suggesting removing _support_ for a property, but to move them to their own copying lines like 
```javascript
if (hasValue(layerObject.popupTemplate)) {
    imageryLayer.popupTemplate = buildJsPopupTemplate(layerObject.popupTemplate);
}
```
this will work, but `copyValues...` does not do any of the careful transfer that `buildJs...` does. So _all_ properties that are not `string`, `number`/`double`/`int`, `bool`, or possibly an array of one of those types, should be removed from this list _and_ implemented separately, usually with a `buildJs...` method. If you think there are properties you want to exclude from the current implementation, then they should be removed from both C# and Typescript completely, and probably added to a separate gissue so we don't forget about them.

TimPurdum (changes requested):
Much cleaner, a bit more work needed to unwind `copyValuesIfExist`.

src/dymaptic.GeoBlazor.Core/Components/Renderers/RasterStretchRenderer.cs
seahro: an oversight, corrected

src/dymaptic.GeoBlazor.Core/Components/Renderers/ImageryRenderer.cs
seahro: corrected and comment added to help assist in understanding renderer inheritance

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.cs
seahro: Added the Prgma warning to: `ImageryLayer.cs, OpenStreetMapLayer.cs, KMLLayer.cs, GraphicsLayer.cs, GeoRSSLayer.cs, GeoJSONLayer.cs and FeatureLayer.cs`

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
seahro: I did break down this list of properties.  I did identify 18 could be built out in a future gissue which creates a number of these builders that can be used by a number of different layers, etc in the future to add depth to the data sets.  As of now, I think it that many just have not been built out because the different layer data that some of these things require.  Further testing with the other imagry renderers in pro may shine some more light on if there are copy values that are failing to work?

TimPurdum: So, this seems to still need work. For each value in the list:
- if it works, leave it. If it is a simple type as described above, you can assume it works and leave it.
- If it doesn't work, remove it, then
  - Start a new gissue for "Adding Property support to ImageryLayer"
  OR
  - Implement copying the value with an `if (hasValue())` pattern and potentially a `buildJs...` call

Before we merge this branch, every property should be either added to a separate gissue, or implemented in a functioning copy.
Just looking at the https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-ImageryLayer.html docs, the following properties probably don't belong in this `copyValues` and need to be dealt with:
- capabilities
- customParameters
- effect
- fields
- fieldsIndex
- mosaicRule
- multidimensionalInfo
- multidimensiohalSubset
- rasterFunction
- rasterFunctionInfos
- serviceRasterInfo
- sourceJSON
- timeExtent
- timeInfo
- timeOffset

Also, there are several properties that are not in this list _and_ not implemented separately, but are important
- renderer
- popupTemplate
- portalItem
- spatialReference

Let me know if you get stuck on implementing any of these.

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
```diff
@@ -23,6 +23,7 @@
         <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 	    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	    <UserSecretsId>9ebf8cd4-ec74-46b1-95be-809da8cfe903</UserSecretsId>
```
TimPurdum, line 26: we should not need user secrets in a class library, only in executable projects.

TimPurdum (changes requested):

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ImageryLayers.razor
```diff
@@ -0,0 +1,53 @@
+﻿@page "/imagerylayer"
```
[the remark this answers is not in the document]
seahro: resolved, Thanks for your help. @AndersenBell

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
seahro: In the above list the gissues were created for:
rasterFunction, MosaicRule, PixelFilter, SpatialReference, TimeInfo, RasterInfo+RasterBandInfo (is the type of serviceRasterInfo), and SourceJSON
Properties included in this PR: multidimensionalInfo, multidimensionalSubset + sub properties

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
seahro: removed

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: If these properties are not currently supported because they have more work to be done, they should not be in the C# class `ImageryLayer` on this commit, or should at least be commented out. Currently, you still have it in a state where to the consumer, these look like they should work, but maybe don't?

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/ImageryLayerTests.razor
```diff
@@ -0,0 +1,86 @@
...
+        <Map>
+            <ImageryLayer @ref="layer" 
+                          Url="https://sampleserver6.arcgisonline.com/arcgis/rest/services/NLCDLandCover2001/ImageServer" />
+        </Map>
+    </MapView>);
+        await WaitForMapToRender();
+        await AssertJavaScript("assertLayerExists", args: "imagery");
+        await AssertJavaScript("assertObjectHasPropertyWithValue", nameof(TestCanRenderImageryLayerCheckAFullExtentValue),
+            layer!.Id, "fullExtent.xmax", 2266845);
+
```
TimPurdum, line 84: you don't have to write a new test for every property you want to verify. You could set a bunch of properties in one test and assert each one in turn.

TimPurdum (changes requested):
1. Add a unit test `AssertJavaScript` for every property you believe should be working.
2. Remove properties that are not complete/working from `ImageryLayer.cs`

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
seahro: Remaining ImageryLayer Class properties were commented out pending completion of the gissues associated with the ImageryLayer and/or other gissues that would enable the use of these particular properties.  The Time related properties have only been partially completed.

TimPurdum (approved):
SHIP IT

PR #226, opened 2023-09-12
Release Merge

TimPurdum:
- Update links in Samples home page
- Added a new sample for clicking to add a point
- Fixes for popup templates and popup actions bugs like #222


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/About.razor.css
```diff
@@ -49,6 +49,16 @@
...
 
+.pro {
+    color: gold;
+    font-style: italic;
+    font-weight: bold;
+    padding: 1rem 0;
+}
+
+.pro a {
+    color: gold!important;
```
agnesstelmach, line 60: just poking in to say generally speaking, it's best not to use !important most of the time in CSS. Specifying the class and element like you did here should be enough to ensure no other rules in this file take priority. Is there maybe another reason !important is needed?

src/dymaptic.GeoBlazor.Core/package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "dymaptic.GeoBlazor.Core",
-  "version": "2.3.0",
+  "version": "2.3.1-beta-1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "dymaptic.GeoBlazor.Core",
-      "version": "2.3.0",
+      "version": "2.3.1-beta-1",
```
AndersenBell, line 9: what is the difference between 2.3.1-beta-1 and everywhere else that has 2.3.2?

AndersenBell (approved):

src/dymaptic.GeoBlazor.Core/package-lock.json
TimPurdum: package-lock.json should generate on build from package.json, guess I forgot to do a build here, will correct before merging!

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/About.razor.css
TimPurdum: @agnesstelmach you are absolutely correct, and it was a shortcut months ago that I used instead of taking the time to figure out why it didn't work without it. I will try to revisit soon, or feel free when you're working in GB to update!

PR #228, opened 2023-09-14
Bug/227 MapView SetZoom

numv:


TimPurdum (approved):
Nice find! That's a downside of those `param object[]` overloads, in that it doesn't enforce the types and tell you when you mis-order them!

I'll get this merged into `main` and a new release this week. Thanks for doing the legwork!

PR #230, opened 2023-09-14
Feature/117 update feataure layer

### LINKED ISSUE #117: Add other methods to FeatureLayer

AndersenBell:
closes #117 

This does not have Save or SaveAs, they need additional testing, so a new gissue will need to be created for it.

This branch fixes a bug with Query on the feature layer, where sending `undefined` is expected by the JS but we were creating an empty query causing unexpected results.

Making the asset copy check if directories exist before copying files

Adding support for feature layer additional functionality

### DIFF
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/CSVLayers.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/CSVLayers.razor
@@ -14,7 +14,7 @@
 <button disabled="@(!_mapRendered)" @onclick="AddLayer">Add new CSV Layer!</button>
 
 
-<SceneView @ref="_view" class="map-view" OnMapRendered="OnMapRendered">
+<SceneView @ref="_view" class="map-view" OnViewRendered="OnMapRendered">
     <Extent Xmin="-41525513" Ymin="4969181" Xmax="-36687355" Ymax="9024624">
         <SpatialReference Wkid="102100" />
     </Extent>
--- a/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
@@ -323,28 +323,6 @@ public class FeatureLayer : Layer
             View!.Id);
     }
 
-    //TODO: needs more testing
-    /// <summary>
-    ///     Saves the layer to its existing portal item in the Portal authenticated within the user's current session. If the layer is not saved to a PortalItem, then you should use saveAs.
-    ///     Known Limitations:
-    ///     You can only save service-backed FeatureLayers. Client-side layers cannot be saved.
-    /// </summary>
-    public async Task Save(bool? ignoreUnsupported)
-    {
-        await JsLayerReference!.InvokeVoidAsync("save", ignoreUnsupported);
-    }
-
-    //TODO: needs more testing
-    /// <summary>
-    ///     Saves the layer to its existing portal item in the Portal authenticated within the user's current session. If the layer is not saved to a PortalItem, then you should use saveAs.
-    ///     Known Limitations:
-    ///     You can only save service-backed FeatureLayers. Client-side layers cannot be saved.
-    /// </summary>
-    public async Task<PortalItem> SaveAs(PortalItem portalItem, PortalFolder? portalFolder, bool? ignoreUnsupported)
-    {
-        return await JsLayerReference!.InvokeAsync<PortalItem>("saveAs", portalItem, portalFolder, ignoreUnsupported);
-    }
-
     /// <summary>
     /// Returns a FeatureType describing the feature's type. This is applicable if the layer containing the feature has a typeIdField.
     /// </summary>
--- a/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
@@ -5,6 +5,7 @@ using dymaptic.GeoBlazor.Core.Components.Widgets;
 using dymaptic.GeoBlazor.Core.Exceptions;
 using dymaptic.GeoBlazor.Core.Objects;
 using Microsoft.AspNetCore.Components;
+using Microsoft.Extensions.Options;
 using Microsoft.JSInterop;
 using System.Text;
 using System.Text.Json;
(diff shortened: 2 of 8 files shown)


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/CSVLayers.razor
```diff
@@ -14,7 +14,7 @@
 <button disabled="@(!_mapRendered)" @onclick="AddLayer">Add new CSV Layer!</button>
 
 
-<SceneView @ref="_view" class="map-view" OnMapRendered="OnMapRendered">
+<SceneView @ref="_view" class="map-view" OnViewRendered="OnMapRendered">
```
seahro, line 17: Do you think it will be good to adjust the other razor pages that use `SceneView` elements to also switch from `OnMapRendered` to `OnViewRendered` for clarity/consistency?  Like the `SceneView` in `ReactiveUtils.razor`.

seahro (approved):
Looks good! One question about `OnViewRendered`

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/CSVLayers.razor
TimPurdum: `OnMapRendered` is obsolete, we should update all our samples to `OnViewRendered` (that is irrespective of `MapView` vs. `SceneView`, they're both `Views`).

src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
```diff
@@ -5,6 +5,7 @@
 using dymaptic.GeoBlazor.Core.Exceptions;
 using dymaptic.GeoBlazor.Core.Objects;
 using Microsoft.AspNetCore.Components;
+using Microsoft.Extensions.Options;
```
TimPurdum, line 8: this doesn't seem related to other changes, could be removed?

src/dymaptic.GeoBlazor.Core/Components/Widgets/Domain.cs
```diff
@@ -77,19 +76,29 @@ public override async Task UnregisterChildComponent(MapComponent child)
...
     /// <summary>
     ///     The value of the code.
     /// </summary>
     [Parameter]
-    public string? Code { get; set; }
+    public T? Code { get; set; }
+}
+
+/// <summary>
+/// Inherited domains apply to domains at the feature type level. It implies that the domain for a given field at the feature
```
TimPurdum, line 95: I realize this is very nitpicky, but can we use the indented XML style I have above for all XML comments?

src/dymaptic.GeoBlazor.Core/Components/PortalFolder.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Microsoft.AspNetCore.Components;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace dymaptic.GeoBlazor.Core.Components;
+public class PortalFolder : MapComponent
```
TimPurdum, line 9: needs class-level xml comments

TimPurdum: Is this class used? I can't find any references.

src/dymaptic.GeoBlazor.Core/Components/PortalFolder.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// The date the folder was created.
+    /// </summary>
+    public DateTime CreatedDate { get; set; }
+
+
+    /// <summary>
+    ///  The portal associated with the folder.
+    /// </summary>
+    public Portal Portal { get; set; } = default!;
```
TimPurdum, line 26: As a `MapComponent`, this would be assumed to work with nested components, so simple properties should be `[Parameter]` and `Portal` would be added via `RegisterChildComponents`.

src/dymaptic.GeoBlazor.Core/Scripts/dotNetBuilder.ts
```diff
@@ -749,3 +752,29 @@ export function buildDotNetTimeInterval(interval: any): any | null {
...
+            dotNetDomains[domain] =  buildDotNetDomain(result.domains[domain]);
+        }
+    }
+
+    return {
+        id: result.id,
+        domains: dotNetDomains,
+        declaredClass: result.declaredClass,
+        name: result.name,
+        templates: result.templates,
```
TimPurdum, line 774: have you validated that `templates` converts OK without a builder method?

src/dymaptic.GeoBlazor.Core/Scripts/featureLayer.ts
```diff
@@ -235,7 +255,64 @@ export default class FeatureLayerWrapper {
...
+
+        return buildDotNetFeatureType(result);
+    }
+
+    async getField(fieldName: string): Promise<DotNetField | null> {
+
+        let result = await this.layer.getField(fieldName);
+
+        if (result != undefined) {
+            let field = buildDotNetFields(Array(result));
```
TimPurdum, line 276: this `Array(result)` could also be `[ result ]`, right? (I'm just more familiar with that pattern, this is fine).

TimPurdum (approved):
Looks good! A few cleanup items.

src/dymaptic.GeoBlazor.Core/Components/PortalFolder.cs
AndersenBell: I think it is referenced when I was trying to get some things workings, but I removed it

src/dymaptic.GeoBlazor.Core/Scripts/featureLayer.ts
AndersenBell: There are some differences, so for consistency, I will use [ ]

src/dymaptic.GeoBlazor.Core/Scripts/dotNetBuilder.ts
AndersenBell: yes

PR #234, opened 2023-09-30
Implemented all properties on TextSymbol, unit tests and Label sample

### LINKED ISSUE #231: TextSymbol Offset
Implementation for offsets in the TextSymbol aren't implemented yet.
SimpleMarkerSymbol already has the offset, but offsetting the exact Dot is not optimal.
Would be better if you could offset the Text underneath the Dot.

taken from arcgis developer portal
```js
let textSymbol = {
  type: "text",
  color: "white",
  haloColor: "black",
  haloSize: "1px",
  text: "You are here",
  xoffset: 3,
  yoffset: 3,
};
```

TimPurdum:
Closes #231 
- Implements missing properties on TextSymbol, including the requested `XOffset` and `YOffset`
- Wrote new `SymbolsTests` to verify that symbol is generated and all properties are set
- Added `Labels` sample that shows off some TextSymbol features

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Symbols/TextSymbol.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Symbols/TextSymbol.cs
@@ -1,4 +1,5 @@
 ﻿using dymaptic.GeoBlazor.Core.Objects;
+using dymaptic.GeoBlazor.Core.Serialization;
 using Microsoft.AspNetCore.Components;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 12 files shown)


seahro (approved):

src/dymaptic.GeoBlazor.Core/Components/Symbols/TextSymbol.cs
```diff
@@ -71,13 +112,93 @@ public TextSymbol()
...
+    
+    /// <summary>
+    ///     The offset on the x-axis in points. This value is a string expressing size in points or pixels (e.g. "12px", "12pt"), which defaults to points.
+    /// </summary>
+    /// <remarks>
+    /// This property is currently not supported in 3D SceneViews.
+    /// </remarks>
+    [Parameter]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? XOffset { get; set; }
```
seahro, line 191: In looking at the `TextSymbol` docs, I noticed that a number of these new properties are autocast, are not supported in 3d Sceneviews or are only applicable when `TextSymbol` is not used for labeling, is this why they are not added in the main `TextSymbol` constructor?

seahro (approved):
looks good.  One comment on TextSymbol class

src/dymaptic.GeoBlazor.Core/Components/Symbols/TextSymbol.cs
TimPurdum: No, they aren't added to the constructor because I forgot to update the constructor 🙂

PR #255, opened 2023-11-01
Feature/advanced feature reductions pro 58

TimPurdum:
- Support for GB Pro features `Feature Reduction Clustering` and `Pie Chart Renderer`
- Added `Effect` property and `SetEffect` method to `FeatureLayer`
- Added `SetSelectedClusterBoundaryFeatureSymbol
- Added `NumberToStringConverter` to deserialize `string | number` properties as a string
- Fixed bug in `DisplayProjection` sample, when loading via Pro
- Fixed bug in `MapComponent.AwaitReactiveSingleWatchUpdate` throwing an exception in JS


src/dymaptic.GeoBlazor.Core/Components/Renderers/Renderer.cs
```diff
@@ -70,6 +90,18 @@ public enum RendererType
...
 #pragma warning disable CS1591
     Simple,
-    UniqueValue
+    UniqueValue,
+    [LookupType("dymaptic.GeoBlazor.Pro.Components.Renderers.PieChartRenderer")]
+    PieChart
 #pragma warning restore CS1591
+}
+
+public class LookupTypeAttribute : Attribute
```
seahro, line 99: Are we going to want a similar pattern created in the "custom" imageryrenderer class that is being setup?

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Bookmarks.razor
```diff
@@ -12,7 +12,7 @@
     Sample demonstrates the use of the Bookmarks widget and Bookmark actions.
     This follows the ESRI sample, but the Timeline Widget is included with GeoBlazor pro.
 </p>
-<MapView @ref="_mapView" Class="map-view" OnLayerViewCreate="OnLayerViewCreate">
+<MapView Class="map-view" OnLayerViewCreate="OnLayerViewCreate">
```
seahro, line 15: I see a number of the samples pages will no longer using the MapView, is this a refactor that we will want to do across most, if not all layer types going forward? For Consistency

seahro (approved):
Approved, couple comments about guidance with ongoing or future work

src/dymaptic.GeoBlazor.Core/Components/Renderers/Renderer.cs
TimPurdum: This is only necessary because of needing the full type of GB Pro classes, and the fact that the `Renderer.Type` was already implemented as an enum and not a string. Most of our `Type` properties are strings, which make this easier.

src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
```diff
@@ -405,6 +412,17 @@ public override void Refresh()
...
         base.Refresh();
     }
+    
+    /// <summary>
+    ///     Effect provides various filter functions that can be performed on the layer to achieve different visual effects similar to how image filters work. This powerful capability allows you to apply css filter-like functions to layers to create custom visual effects to enhance the cartographic quality of your maps. This is done by applying the desired effect to the layer's effect property as a string or an array of objects to set scale dependent effects.
+    /// </summary>
+    /// <param name="effect">
+    ///     The effect to apply to the layer.
+    /// </param>
+    public async Task SetEffect(Effect effect)
```
AndersenBell, line 422: is there a way to remove an effect?

src/dymaptic.GeoBlazor.Core/Components/Renderers/Renderer.cs
AndersenBell: should there be a TODO to switch this to a string for the next major release then?

AndersenBell (approved):
looks good

AndersenBell (approved):
looks good

src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
TimPurdum: I will make a gissue for this.

PR #263, opened 2023-11-15
fix pro wasm bug, refactor JS module loading

### LINKED ISSUE #262: Refactor JS Module Loading to fix WASM Sync-JS bug in GeoBlazor Pro
Discovered that the WASM `AddGraphics` implementation using the synchronous `JSHost` was broken in GeoBlazor Pro, due to an inconsistent JS module loading pattern.

TimPurdum:
- Moved all JS Module loading into `JsModuleManager` static class
- Changed the order of module loading so that Pro never gets loaded twice
- Fix the bug, closes #262 , where Pro was not loading the Core JS via the `Pro.getCore` method. I did this by loading the pro in the `JSHost`, and then having a duplicated implementation on the JS side of the synchronous method.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
@@ -1414,9 +1414,24 @@ public partial class MapView : MapComponent
     }
 
 #if NET7_0_OR_GREATER
+    internal void AddGraphicsSyncInterop(byte[] graphics, string id, string? layerId = null)
+    {
+        if (ProJsViewModule is not null)
+        {
+            AddGraphicsProSyncInterop(graphics, id, layerId);
+        }
+        else
+        {
+            AddGraphicsCoreSyncInterop(graphics, id, layerId);
+        }
+    }
+
 #pragma warning disable CA1416
-    [JSImport("addGraphicsSyncInterop", "arcGisJsInterop")]
-    internal static partial void AddGraphicsSyncInterop(byte[] graphics, string id, string? layerId = null);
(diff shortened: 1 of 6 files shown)


src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -1414,9 +1414,24 @@ public async Task AddGraphics(IEnumerable<Graphic> graphics, CancellationToken c
     }
 
 #if NET7_0_OR_GREATER
+    internal void AddGraphicsSyncInterop(byte[] graphics, string id, string? layerId = null)
+    {
+        if (ProJsViewModule is not null)
+        {
+            AddGraphicsProSyncInterop(graphics, id, layerId);
```
seahro, line 1421: I assume the pro code PR will have this method adjustment, since the `AddGraphicsCoreSyncInterop` is the only method in Core's `arcGISJsInterop.ts`?

seahro (approved):
one comment.  Looks good.  There have been multiple refactors to switch `MapView` to `View`.  Will need to keep an eye on that in the other open PRs.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: Yes, the method didn't actually exist at all in `arcGisPro.ts` yet, it is added in that PR.

TimPurdum:
> one comment. Looks good. There have been multiple refactors to switch `MapView` to `View`. Will need to keep an eye on that in the other open PRs.

This was actually switching from a static method to an instance method. `MapView.AddSync...` was a static method, so it was on the class type, not on the instance, which was stored as `View` in `Layer.cs`.

PR #269, opened 2023-11-30
adding/updating javascript file versioning

### LINKED ISSUE #268: Version JavaScript files to avoid caching issues
We should add a versioned query param to the loading of the JavaScript modules to avoid issues with browsers caching an older version when an application is updated.

e.g. in `JsModuleManager`

```csharp
return await jsRuntime
                    .InvokeAsync<IJSObjectReference>("import", cancellationToken, 
                        "./_content/dymaptic.GeoBlazor.Core/js/arcGisJsInterop.js?v=2.5.0");
```

The tricky part would be to set that version in code based on the `Directory.Build.props` environment variable `CoreVersion`.

AndersenBell:
closes #268

Adds a query parameter that is the current GeoBlazor version to the requests for the js interop file to prevent browser caching after an update.

Also combined duplicate interop request code to a single source.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/JsModuleManager.cs
+++ b/src/dymaptic.GeoBlazor.Core/JsModuleManager.cs
@@ -1,4 +1,5 @@
 using Microsoft.JSInterop;
+using System.Security.Claims;
 
 namespace dymaptic.GeoBlazor.Core;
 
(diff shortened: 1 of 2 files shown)


seahro (approved):
Looks good.

TimPurdum (approved):
Nice! Thanks for consolidating the calls as well.

PR #271, opened 2023-12-02
removed package reference, rebuilt and tested

### LINKED ISSUE #270: Clean up unused and warning-causing NPM reference
I initially thought that the warnings in NPM packages and Dependabot were coming via ArcGIS transient references. However, it turns out they are instead references from protobufjs-cli, which in turn apparently is not needed, I had imported it while investigating protobuf graphics serialization.

Remove `protobufjs-cli` reference and test.

TimPurdum:
Closes #270

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/package-lock.json
+++ b/src/dymaptic.GeoBlazor.Core/package-lock.json
@@ -8,19 +8,18 @@
       "name": "dymaptic.GeoBlazor.Core",
       "license": "ISC",
       "dependencies": {
-        "@arcgis/core": "4.28",
-        "esbuild": "0.19.5",
-        "protobufjs": "7.2.5",
-        "protobufjs-cli": "1.1.2"
+        "@arcgis/core": "4.28.10",
+        "esbuild": "0.19.8",
+        "protobufjs": "7.2.5"
       },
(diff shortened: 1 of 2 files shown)


src/dymaptic.GeoBlazor.Core/package-lock.json
```diff
@@ -8,19 +8,18 @@
       "name": "dymaptic.GeoBlazor.Core",
       "license": "ISC",
       "dependencies": {
-        "@arcgis/core": "4.28",
-        "esbuild": "0.19.5",
-        "protobufjs": "7.2.5",
-        "protobufjs-cli": "1.1.2"
+        "@arcgis/core": "4.28.10",
+        "esbuild": "0.19.8",
+        "protobufjs": "7.2.5"
```
seahro, line 13: so we are just removing the cli? esbuild and arcgis are the updates.

seahro (approved):
Looks good, and the arc version was updated.

src/dymaptic.GeoBlazor.Core/package-lock.json
TimPurdum: protobufjs-cli, yes.

PR #277, opened 2023-12-16
FeatureLayer Client-side Graphics method fixes

### LINKED ISSUE #275: Cannot Create an Empty FeatureLayer for Client-side Graphics
GeoBlazor currently enforces that you do one of the following when declaring a `FeatureLayer`:
   - Define a `Url`
   - Define a `PortalItem`
   - Add one or more child `Graphics` in markup, which creates the `Source` object for local graphics.

The problem arises if you want to define the FeatureLayer in markup, but not add any graphics until later with `ApplyEdits`. For this pattern, we need to allow users to define `Source` as an attribute with an empty array.

We also should update documentation around this, and the fact that the layer will not _render_ in ArcGIS until there are graphics, which means you also need to call `layer.Load()` before `layer.ApplyEdits()`.

TimPurdum:
Closes #275 

- You can now define an empty `Source` parameter like below (C#12 syntax). This makes it possible to start with an empty layer and then add graphics later programmatically.

```html
<FeatureLayer Source="@([])">
```

- Discovered that an empty layer like above never calls `layerview-create`, and thus isn't "registered" in GeoBlazor with a JSObjectReference. Fixed `ApplyEdits` to also call `Load` first if the object ref isn't set.
- Added specific thrown exceptions to `FeatureLayer.Add(Graphic graphic)` and  `FeatureLayer.Remove(Graphic graphic)` methods if these are called after the layer is rendered, because ArcGIS Feature Layers do not support updating graphics this way.
- Did some cleanup on `ImageryLayer` and the related renderers to remove broken code that was either not converting correctly in TypeScript, or was trying to set read-only fields
- Changed `Label.LabelPlacement` to be an enum instead of a string. While this is technically a breaking change, I did discover a bug when the value would round-trip from ArcGIS, where ArcGIS returned values like `esriServerPointLabelPlacementAboveCenter`, so this also fixes that bug.
- Removed broken Pro code in `Layer.Load`
- Updated `testRunner.js` to use a better injection pattern for extending GeoBlazor.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Serialization/EnumToKebabCaseStringConverter.cs
+++ b/src/dymaptic.GeoBlazor.Core/Serialization/EnumToKebabCaseStringConverter.cs
@@ -1,4 +1,5 @@
-﻿using dymaptic.GeoBlazor.Core.Extensions;
+﻿using dymaptic.GeoBlazor.Core.Components.Layers;
+using dymaptic.GeoBlazor.Core.Extensions;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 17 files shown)


src/dymaptic.GeoBlazor.Core/Serialization/EnumToKebabCaseStringConverter.cs
```diff
@@ -68,4 +69,29 @@ public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerial
...
+            ?.Replace("-", string.Empty)
+            .Replace("esriServerPointLabelPlacement", string.Empty);
+
+        try
+        {
+            return value is not null ? (LabelPlacement)Enum.Parse(typeof(LabelPlacement), value, true) : default(LabelPlacement)!;
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine(ex);
```
AndersenBell, line 92: should this be a console log?

TimPurdum: In general, I've never hooked up GeoBlazor to ILogger, although I suppose we could/should. Since it's a library, I wasn't sure how I felt about "pushing" log messages to someone's logger implementation? Anyways, that seems like it should be it's own library-wide gissue. Let's talk about it more.

PR #278, opened 2023-12-18
Fix for snapping back to param-defined extent on mouse wheel

### LINKED ISSUE #276: MapView Zoom Resets when OnViewRendered Defined on View
The zoom value seems to reset when trying to pinch zoom or use zoom controls when the mapview first loads.

Reproduce Issue

- Clone sample repo here: https://github.com/djrpascu/GeoBlazorBugMapViewZoomReset
- Wait for map to load, without panning, try zoom in/out with control, or pinch zoom, the map zoom should reset.

Similar to the issue reported here:
https://github.com/dymaptic/GeoBlazor/issues/188
@TimPurdum 

- Like in the issue above, if you pan the map first, then try zoom controls, the zoom is retained.

- Remove `OnViewRendered="OnViewRendered"` from MapView in MyMapComponent.razor. Try again, and should work.

TimPurdum:
Closes #276  

Added a second layer of "defense" against user interactions on map being "overridden" by properties re-syncing in Blazor. This happens in `argGisJsInterop` by setting a boolean when the user interacts via event handlers.

Also did some reorganization of `MapView.razor.cs`, and started some investigations into what appears to be a GeoBlazor Pro-only bug affecting popup Actions.


src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -2249,80 +2112,204 @@ protected override async Task RenderView(bool forceRender = false)
...
+    
+    private bool IsPro()
+    {
+        if (_isPro is null)
         {
             try
             {
-                await ViewJsModule!.InvokeVoidAsync("removeWidget",
-                    CancellationTokenSource.Token, widget.Id, Id);
+                Assembly _ = Assembly.Load("dymaptic.GeoBlazor.Pro");
```
AndersenBell, line 2300: is there a better way to check then throwing an exception?

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -2655,6 +2664,11 @@ function waitForRender(viewId: string, dotNetRef: any): void {
     })
 }
 
+function setUserChangedViewExtent() {
+    console.log('user changed view extent');
```
AndersenBell, line 2668: do you want this .log statement?

AndersenBell (approved):

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
TimPurdum: no, good catch

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: I can't find it. There is `AppDomain.CurrentDomain.GetAssemblies()` which you can iterate through, but that says it gets "loaded" assemblies, and I'm not sure if it would be loaded at this point before calling `Assembly.Load`.

PR #279, opened 2023-12-19
Release Merge

TimPurdum:
# Release Notes

## New Features
- Added Graphic.Visible parameter

## Bug Fixes
- Fix for starting a client-side `FeatureLayer` with an empty `Source` #275 
- Fix for mouse-wheel on first navigation snapping back to parameter-set `Extent` #276 
- Fix Guid syncing in Graphic attributes
- Fix for `View.ToScreen` and `View.HitTest` point calculations bug
- Fix Calcite versions for Sample apps
- Fix bug with `Label.LabelPlacement` (see Breaking Change below)
- Deprecate/throw exception for `FeatureLayer.Add(graphic)`

## Breaking Change
- Due to a serialization bug, we had to replace `Label.LabelPlacement` with an enum


src/dymaptic.GeoBlazor.Core/Components/Layers/Graphic.cs
```diff
@@ -68,6 +72,13 @@ public Graphic()
     [Parameter]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public AttributesDictionary Attributes { get; set; } = new();
+    
+    /// <summary>
+    ///     Indicates the visibility of the graphic. Default value: true.
+    /// </summary>
+    [Parameter]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public bool? Visible { get; set; }
```
AndersenBell, line 81: so it default to true in JS if its null?

TimPurdum: Yes, using the `copyValuesIfExist` pattern in the TypeScript, we only set this if the user does, otherwise we leave it unset and the ArcGIS defaults take over.

PR #282, opened 2023-12-27
Implemented and tested ApplyEdits changes via protobuf serialization

### LINKED ISSUE #281: FeatureLayer AddEdits support for large graphic sets
Currently, GraphicsLayers are built to support large sets of graphics by using Protobuf serialization and data chunking. However, this was not implemented with FeatureLayer.ApplyEdits, which still uses JSON serialization and no chunking.

Update FeatureLayer.ApplyEdits to use the protobuf serialization and chunking to support large data sets.

TimPurdum:
Closes #281 

- Implemented chunking and protobuf serialization of graphics in `FeatureLayer.ApplyEdits`
- Wrote unit tests for `ApplyEdits`
- Also re-implemented `FeatureLayer.Add` so that it uses `ApplyEdits` if the layer is already rendered.

### DIFF
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ManyGraphics.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ManyGraphics.razor
@@ -1,7 +1,7 @@
 ﻿@page "/many-graphics"
 
 <PageTitle>Many Graphics</PageTitle>
-<h1>Many Points</h1>
+<h1>Many Graphics</h1>
 <div class="links-div">
     <a class="btn btn-secondary" target="_blank" href="https://developers.arcgis.com/javascript/latest/api-reference/esri-Graphic.html">ArcGIS API for JavaScript</a>
 </div>
--- a/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
@@ -8,6 +8,7 @@ using dymaptic.GeoBlazor.Core.Objects;
 using dymaptic.GeoBlazor.Core.Serialization;
 using Microsoft.AspNetCore.Components;
 using Microsoft.JSInterop;
+using ProtoBuf;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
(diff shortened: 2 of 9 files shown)


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ManyGraphics.razor
```diff
@@ -1,7 +1,7 @@
 ﻿@page "/many-graphics"
 
 <PageTitle>Many Graphics</PageTitle>
-<h1>Many Points</h1>
+<h1>Many Graphics</h1>
 <div class="links-div">
     <a class="btn btn-secondary" target="_blank" href="https://developers.arcgis.com/javascript/latest/api-reference/esri-Graphic.html">ArcGIS API for JavaScript</a>
 </div>
```
seahro, line 7: maybe add a quick blurb referencing the one exception for feature layers if whether it is loaded or not. "If the layer is already loaded, you must use <see cref="ApplyEdits"/> to add graphics."

seahro (approved):
Looks good.  One comment on the item of note that if the layer is already loaded, you need to use ApplyEdits...?

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ManyGraphics.razor
TimPurdum: In this PR I actually fixed that, I believe. Now `Add` uses `ApplyEdits` under the hood and doesn't fail.

src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
```diff
@@ -294,13 +295,18 @@ public FeatureLayer()
     /// <exception cref="InvalidOperationException">
     ///     If the layer is already loaded, you must use <see cref="ApplyEdits"/> to add graphics.
     /// </exception>
-    public Task Add(Graphic graphic)
+    public async Task Add(Graphic graphic)
     {
         if (JsLayerReference is not null)
         {
-            throw new InvalidOperationException("Cannot add graphics to a feature layer that is already loaded. Use ApplyEdits instead.");
+            await ApplyEdits(new FeatureEdits
```
TimPurdum, line 302: @seahro here you can see where I removed the exception and re-implemented using `ApplyEdits`

PR #283, opened 2023-12-29
Graphics Legend Widget

TimPurdum:
- Support for GeoBlazor Pro Issue 91, `GraphicsLegendWidget`
- Consolidates `Visible` as a property on `MapComponent`, and removes from some inherited classes.


seahro (approved):
Moving Visibility to the map component level makes sense since so many items inherit from it.  Excited to see the Pro GraphicsLegendWidget.

seahro (approved):
Moving Visibility to the map component level makes sense since so many items inherit from it.  Excited to see the Pro GraphicsLegendWidget.

PR #287, opened 2024-01-07
Version/3 dimensional properties

TimPurdum:
This is the start of dedicated V3.0 changes for GeoBlazor. Each PR will feed into the `version/3.0.0` branch, which will also be kept up to date from `develop`. Then, when we are ready to release 3.0, we will merge that back into develop and main.

In this PR
- Removed methods marked `obsolete` in previous releases
- Replaced inconsistent `number | string` TS properties for pixel/point sizes with a new class `Dimension`


src/dymaptic.GeoBlazor.Core/Extensions/Dimension.cs
```diff
@@ -0,0 +1,139 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+
+namespace dymaptic.GeoBlazor.Core.Extensions;
+
+/// <summary>
+///     Defines a screen measurement in points or pixels. Supports implicit conversion from string, int, or double.
```
seahro, line 8: There are a number of Dimension references in the JS API.  Confirming this dimension extension class lays the groundwork for the properties found in the DimensionSimpleStyle correct? https://developers.arcgis.com/javascript/latest/api-reference/esri-analysis-DimensionSimpleStyle.html

seahro (approved):
Looks good.  I had one question that I think I answered correctly.  See Dimension.CS

src/dymaptic.GeoBlazor.Core/Components/Symbols/Symbol.cs
```diff
@@ -131,7 +132,7 @@ public SymbolSerializationRecord()
     public MapColor? HaloColor { get; init; }
 
     [ProtoMember(13)]
-    public int? HaloSize { get; init; }
+    public double? HaloSize { get; init; }
```
AndersenBell, line 135: is there a reason these are doubles, but in TextSymbol they are Dimension?

AndersenBell (approved):

src/dymaptic.GeoBlazor.Core/Extensions/Dimension.cs
TimPurdum: Esri's use of the word `Dimension` in `DimensionSimpleStyle` and `DimensionAnalysis` seem to be focused on real-world dimensions, whereas this is for screen measurements. However, since they aren't using the exact same word as a type (so far), I think it's fine and there is no conflict.

src/dymaptic.GeoBlazor.Core/Components/Symbols/Symbol.cs
TimPurdum: Yes, good question. This is a protobuf serialization class, and I decided to handle all the conversions within the C#, not expose them to the serialization or JS.

PR #290, opened 2024-01-13
Basemap reference layers and widget options

### LINKED ISSUE #288: Support Reference Layers in Basemap
Currently, we are adding all layers within the `<Basemap>` tags to `view.map.basemap.baseLayers`. However, for specific `reference` layers, they are intended to be added to a separate property, `view.map.basemap.referenceLayers`.

https://developers.arcgis.com/javascript/latest/api-reference/esri-Basemap.html#referenceLayers

As far as I can tell, this does two things:
1. Separates the layers in the list of the `BasemapLayerListWidget`
2. Moves these reference layers on top of _everything else_, including graphics and graphic layers. This could be intentional depending on your use case, and we should follow the ArcGIS spec.

### LINKED ISSUE #289: Allow BasemapLayerListWidget to toggle Visible Elements
https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-BasemapLayerList.html#visibleElements

BasemapLayerListWidget should have the ability to show or hide the following elements:
- Status Indicators
- Base Layers
- Reference Layers (#288 )
- Errors

TimPurdum:
Closes #288 
-  Added `IsBasemapReferenceLayer` property to `Layer.cs`
Closes #289
- Added `BasemapLayerListVisibleElements`
- Updated sample page with test of toggling sections
  - removed partially implemented callback functions from sample.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
@@ -21,14 +21,6 @@ public class BasemapLayerListWidget : Widget
     public override string WidgetType => "basemapLayerList";
 
     /// <summary>
-    ///     The widget's default CSS icon class.
-    /// </summary>
-    [Parameter]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    [Obsolete("Use Icon instead")]
-    public string? IconClass { get; set; }
-
-    /// <summary>
     ///     The widget's default label.
     /// </summary>
     [Parameter]
--- a/src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
@@ -178,6 +178,10 @@ public class BasemapLayerListWidget : Widget
     }
 }
 
+/// <summary>
+///     The visible elements that are displayed within the widget. This property provides the ability to turn individual elements of the widget's display on/off.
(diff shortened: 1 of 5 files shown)


src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
```diff
@@ -25,6 +25,7 @@ public class BasemapLayerListWidget : Widget
     /// </summary>
     [Parameter]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [Obsolete("Use Icon instead")]
```
AndersenBell, line 28: is this removed in the other PR, can we just remove this here?

src/dymaptic.GeoBlazor.Core/Components/Widgets/LayerListWidget.cs
```diff
@@ -23,6 +23,7 @@ public class LayerListWidget : Widget
     /// </summary>
     [Parameter]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [Obsolete("Use Icon instead")]
```
AndersenBell, line 26: same question, can this be removed completely?

AndersenBell (approved):

src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
TimPurdum: This was originally targeted to update `develop` rather than `v3.0.0`, hence the difference in leaving the `Obsolete` tags. However, after thinking it through, I decided to go ahead and move this with the other new PRs to target v3, and remove the `IconClass`.

PR #291, opened 2024-01-13
Version/3 Reverse order of layers to match expected results

TimPurdum:
Based on an internal discussion, we realized that, currently, if you add layers in GeoBlazor markup, like:

```html
<Map>
    <FeatureLayer Title="Layer 1" />
    <FeatureLayer Title="Layer 2" />
    <FeatureLayer Title="Layer 3" />
    <FeatureLayer Title="Layer 4" />
</Map>
```

these layers will be added to ArcGIS in such a way as they will show in the `LayerListWidget` in reverse order, and more importantly, that Layer 4 will be layered "on top".

This breaks the visual expectation of the markup layout. However, _changing_ the behavior will break anyone who has already implemented stacked layers, so I will also create a dedicated page or section in the docs to explain how it works, and we need to make sure to highlight it in the release notes of V3.

### DIFF
--- a/templates/content/dymaptic.GeoBlazor.Template.WebAssembly/dymaptic.GeoBlazor.Template.WebAssembly.csproj
+++ b/templates/content/dymaptic.GeoBlazor.Template.WebAssembly/dymaptic.GeoBlazor.Template.WebAssembly.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="dymaptic.GeoBlazor.Core" Version="2.5.3" />
+    <PackageReference Include="dymaptic.GeoBlazor.Core" Version="3.0.0-beta-1" />
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="8.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="8.0.0" PrivateAssets="all" />
   </ItemGroup>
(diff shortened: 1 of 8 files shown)


templates/content/dymaptic.GeoBlazor.Template.WebAssembly/dymaptic.GeoBlazor.Template.WebAssembly.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="dymaptic.GeoBlazor.Core" Version="2.5.3" />
```
AndersenBell, line 10: I wonder if there is a way to say use the latest, does * work?

AndersenBell (approved):

AndersenBell:
Is there going to be an additional task for updating the docs, or could that be added to this PR?

templates/content/dymaptic.GeoBlazor.Template.WebAssembly/dymaptic.GeoBlazor.Template.WebAssembly.csproj
TimPurdum: Part of our automated build in github actions automatically updates this when I change the version # in the props file.

TimPurdum:
> Is there going to be an additional task for updating the docs, or could that be added to this PR?

Docs are currently in the `Pro` repository, so I will make a separate gissue for those.

PR #292, opened 2024-01-13
Version/3 Basemap Styles

TimPurdum:
ArcGIS has updated how they define basemaps. They have a new basemap rest service https://developers.arcgis.com/rest/basemap-styles/ that defines the "default" basemaps hosted by Esri with the following JS code:

```js
basemap: {
    style: {
        id: "arcgis/topographic"
    }
}
```

This looks like it will eventually replace being able to set `view.map.basemap` directly to a string value, and the old strings are marked as `legacy` here https://developers.arcgis.com/javascript/latest/api-reference/esri-Map.html#basemap.

To keep up with ArcGIS, and actually make discoverability simpler in GeoBlazor, I'm creating an `enum` that has the entire list of style `id`s. I chose to call them `Name` in GeoBlazor, since that seems to make more sense for a string/enum.

Here is the implementation in GeoBlazor:

```html
<MapView>
    <Map>
        <Basemap>
            <BasemapStyle Name="BasemapStyleName.ArcgisTopographic" />
        </Basemap>
    </Map>
</MapView>
```

I have also
- Marked `Map.ArcGISDefaultBasemap` as `Obsolete`, but not deleting, for a smoother transition.
- Updated the `Basemaps` sample to allow users to test out the new styles enum with a select dropdown.

### DIFF
--- a/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Basemaps.razor
+++ b/samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Basemaps.razor
@@ -9,12 +9,21 @@
 <div class="flex-column">
     <div class="toggle-row">
         <label class="radio-container">
-            From Static Default
+            From Style
             <input type="radio" checked
                    name="basemap-type" @onclick="@(() => SetBasemap(0))" />
             <div class="radio-box" />
         </label>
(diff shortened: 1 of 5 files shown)


morehavoc (commented):
I'm very excited to have this as an Enum for completing all of the codes instead of having to remember these names!

PR #296, opened 2024-02-01
prevent returning all graphics for layer on hitTest.

### LINKED ISSUE #294: Hit Test returning all graphics in layer
The results from a hit test in JavaScript include the following object graph:

results
- graphics
  - graphic
    - layer
      - graphics (all graphics in the layer)

We are serializing all of this, which makes HitTest much slower than expected. Instead, we should only need to return the GeoBlazor Layer Id, and nothing else about the layer.

TimPurdum:
Closes #294

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Scripts/dotNetBuilder.ts
+++ b/src/dymaptic.GeoBlazor.Core/Scripts/dotNetBuilder.ts
@@ -376,12 +376,13 @@ export function buildDotNetLayerView(layerView: LayerView): DotNetLayerView {
     }
 }
 
-export function buildDotNetLayer(layer: Layer): DotNetLayer {
+export function buildDotNetLayer(layer: Layer, includeGraphics: boolean = true)
+    : DotNetLayer {
     switch (layer.type) {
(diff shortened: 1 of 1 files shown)


src/dymaptic.GeoBlazor.Core/Scripts/dotNetBuilder.ts
```diff
@@ -564,10 +567,9 @@ function buildDotNetViewHit(viewHit: ViewHit): DotNetViewHit | null {
             return {
                 type: "graphic",
                 graphic: buildDotNetGraphic(viewHit.graphic),
-                layer: buildDotNetLayer(viewHit.layer ?? viewHit.graphic.layer),
+                layer: buildDotNetLayer(viewHit.layer ?? viewHit.graphic.layer, false),
```
seahro, line 570: setting false on `buildDotNetLayer` will limit the return to just the graphic layer. true would enable the complete layer and all graphics to be returned

seahro (approved):
looks good, adding the boolean makes sense to just return the layer, vs everything

PR #298, opened 2024-02-02
protobuf serialization fixes

### LINKED ISSUE #295: HorizontalAlignment and VerticalAlignment doesn't start with lower case
In namespace dymaptic.GeoBlazor.Core.Components.Symbols;

public enum HorizontalAlignment
{
#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
    left,
    right,
    center
#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
}

public enum VerticalAlignment
{
#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
    baseline,
    top,
    middle,
    bottom
#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
}

Make at example Left to left, because otherwise it will not work

Further in function
    internal override SymbolSerializationRecord ToSerializationRecord()

not exist XOffset and YOffset

TimPurdum:
Closes #295 

- Enums serialized to Protobuf need to be run through `ToKebabCase`, which was missed for `TextSymbol.HorizontalAlignment` and `TextSymbol.VerticalAlignment`
- XOffset and YOffset in `TextSymbol` were not serialized via Protobuf
- `SimpleLineSymbol.LineStyle` was not correctly setting `style` in JS.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Symbols/SimpleLineSymbol.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Symbols/SimpleLineSymbol.cs
@@ -59,7 +59,8 @@ public class SimpleLineSymbol : LineSymbol
     {
         return new SymbolSerializationRecord(Type, Color)
         {
-            Width = Width, LineStyle = LineStyle?.ToString().ToKebabCase()
+            Width = Width, 
+            Style = LineStyle?.ToString().ToKebabCase()
         };
     }
 }
--- a/src/dymaptic.GeoBlazor.Core/Components/Layers/Graphic.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Layers/Graphic.cs
@@ -315,7 +315,7 @@ public class Graphic : LayerObject
                 Geometry?.ToSerializationRecord(),
                 Symbol?.ToSerializationRecord(),
                 PopupTemplate?.ToSerializationRecord(),
-                Attributes?.ToSerializationRecord());
+                Attributes.ToSerializationRecord());
         }
 
         return _serializationRecord;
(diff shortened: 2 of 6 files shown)


src/dymaptic.GeoBlazor.Core/Components/Symbols/SimpleLineSymbol.cs
```diff
@@ -59,7 +59,8 @@ internal override SymbolSerializationRecord ToSerializationRecord()
     {
         return new SymbolSerializationRecord(Type, Color)
         {
-            Width = Width, LineStyle = LineStyle?.ToString().ToKebabCase()
+            Width = Width, 
+            Style = LineStyle?.ToString().ToKebabCase()
```
seahro, line 63: As as above in the `SimpleFillSymbol` and the `FillStyle = FillStyle....` Initially, I read this as it should this change as unnecessary and keep `LineStyle = LineStyle....` to match the parameter above.  However, found the property is 'style' in the docs, so I think that answers my questions.

seahro (approved):
looks good.  I had a comment on the FillStyle and LineStyle...but I believe I answered my question.

src/dymaptic.GeoBlazor.Core/Components/Symbols/SimpleLineSymbol.cs
seahro: One other thing.  Are we missing a couple properties on the the simpleLineSymbol?  Cap, miterLimit, and marker?

TimPurdum: I'm sure we are, but that is a separate issue. I will create one.

PR #301, opened 2024-02-03
Version/3 stream all graphics

TimPurdum:
Moves all identified returning queries of graphics to streaming protobuf serialization
- `FeatureLayer.QueryFeatures`
- `FeatureLayer.QueryTopFeatures`
- `FeatureLayer.QueryRelatedFeatures`
- `FeatureLayerView.QueryFeatures`
- `MapView.HitTest`

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
@@ -879,23 +879,20 @@ public class FeatureLayer : Layer, IFeatureReductionLayer
     public async Task<FeatureSet?> QueryFeatures(Query? query = null, CancellationToken cancellationToken = default)
     {
         IJSObjectReference abortSignal = await AbortManager!.CreateAbortSignal(cancellationToken);
+        Guid queryId = Guid.NewGuid();
 
-        FeatureSet? result = await JsLayerReference!.InvokeAsync<FeatureSet?>("queryFeatures", cancellationToken,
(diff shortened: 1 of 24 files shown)


src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
```diff
@@ -879,23 +880,20 @@ public async Task<int> QueryFeatureCount(Query? query = null, CancellationToken
     public async Task<FeatureSet?> QueryFeatures(Query? query = null, CancellationToken cancellationToken = default)
     {
         IJSObjectReference abortSignal = await AbortManager!.CreateAbortSignal(cancellationToken);
+        Guid queryId = Guid.NewGuid();
 
-        FeatureSet? result = await JsLayerReference!.InvokeAsync<FeatureSet?>("queryFeatures", cancellationToken,
-            query, new { signal = abortSignal }, DotNetObjectReference.Create(this), View?.Id);
+        FeatureSet result = (await JsLayerReference!.InvokeAsync<FeatureSet?>("queryFeatures", cancellationToken,
```
seahro, line 885: are we expecting each `result` to not be null?

seahro (approved):
looks good, one question

src/dymaptic.GeoBlazor.Core/Components/Layers/FeatureLayer.cs
TimPurdum: yes, this should always return something, even if the `FeatureSet` is empty of graphics.

PR #305, opened 2024-02-28
Version/3 map image layer

### LINKED ISSUE #165: Add MapImageLayer
https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-MapImageLayer.html

TimPurdum:
Closes #165 

- Adds `MapImageLayer`
- Adds `Sublayer`
- Adds new `MapImageLayers` sample, with multiple implementations
- Adds `SliderWidget` for one of the samples
- Adds missing constructors for `SimpleRenderer`
- Adds new `LineStyle` options for `SimpleLineRenderer`
- Fixed some mapping and added properties for `ListItem` in `LayerListWidget` and `BasemapLayerListWidget`
- Fixed broken serialization of `GoToOverrideParameters` and `GoToTarget` property for overriding zoom in `LocatorWidget` and `SearchWidget`
- Updated `MapComponent` to support toggling visibility of components in code.
- Added `GetExtentCenter`, `GetExtentHeight`, and `GetExtentWidth` to `GeometryEngine`. In ArcGIS these are properties on `Extent`, but because they are readonly, and an `Extent` might not have a `JsModule`, this is a safter pattern for GeoBlazor


src/dymaptic.GeoBlazor.Core/Components/Widgets/SliderWidget.cs
```diff
@@ -0,0 +1,1104 @@
+using dymaptic.GeoBlazor.Core.Serialization;
+using Microsoft.AspNetCore.Components;
+using Microsoft.JSInterop;
+using System.Text.Json.Serialization;
+
+
+namespace dymaptic.GeoBlazor.Core.Components.Widgets;
+
+/// <summary>
+///     A slider widget that can be used for filtering data, or gathering numeric input from a user. The slider can have multiple thumbs, and provides you with the ability to format labels and control user input.
```
seahro, line 10: I saw the map image page.  Esri docs had a nice slider "parts" diagram, which would lend itself to enabling a user to do some customization.  I dont know if the slider will need its own "demo" page or not.  But, at least it would assist the user understanding the different pieces and what customizations are possible.

seahro (approved):
the slider widget is a nice addition, May want to add notes on some of the customization options see comment `SliderWidget.cs`

src/dymaptic.GeoBlazor.Core/Components/Widgets/SliderWidget.cs
TimPurdum: I don't think we have an easy way to add images to the docs, but I will add what I almost always remember (but forgot here), a link to that page in the Esri docs.

PR #314, opened 2024-03-17
Release v3.0.0

TimPurdum:
# New Features
- `MapImageLayer` - supports loading from `MapService` on ArcGIS portals
- `Sublayer` - specify or inspect the layers in a `MapImageLayer` or `TileLayer`
- `SliderWidget` - control any data with a custom slider
- `PictureFillSymbol`- display custom icons for polygon fill
- `Locator` - tool for geolocating or reverse geolocating addresses
- `MapView.QueryResultsMaxSizeLimit` - controls the limit on streaming graphic results from `HitTest` or layer queries
- `BasemapLayerListWidget` new properties
	- `BasemapTitle`
	- `EditingEnabled`
	- `HeadingLevel`
	- `MultipleSelectionEnabled`
	- `VisibleElements`
- `Basemap.Style` - accesses newer version of ArcGIS and OpenStreetMap basemaps via a `BasemapStyle` enum
	- `ArcGISDefaultBasemap` marked as `Obsolete`, suggested to use `Basemap.Style` instead
- `Dimension`-  custom screen measurement class which can be constructed with either a `string` or a `double` value, and supports `px` and `pt` in the string constructor
- `GeometryEngine` new methods
	- `GetExtentCenter` - returns the center point of an `Extent`
	- `GetExtentHeight` - returns the height of an `Extent` in map units
	- `GetExtentWidth` - returns the width of an `Extent` in map units

# Breaking Changes

- The order of `Layers` added to the map in Razor markup has been inverted. Now, the topmost layer in Razor markup will be on top in the map view. For multi-layered scenarios, you will likely need to reverse the order in your Razor code.
- Converted numerous properties/parameters from `HashSet` to `List` or `IReadOnlyList` to maintain order of items correctly.
- `HitTestResult.GraphicHit` now returns a GeoBlazor `LayerId` rather than a full `Layer` object
- Screen-size measurements now take a custom `Dimension` object
	- `Label.RepeatLabelDistance`
	- `MapFont.Size`
	- `LineSymbol.Width`
	- `MarkerSymbol.XOffset` and `MarkerSymbol.YOffset`
	- `SimpleMarkerSymbol.Size`
	- `TextSymbol.HaloSize`, `TextSymbol.LineWidth`, `TextSymbol.XOffset`, and `TextSymbol.YOffset`
- `SimpleMarkerSymbol.Style` removed, use `MarkerStyle` instead
- `AttributesDictionary` values are now `object?` instead of `object`, and constructor dictionaries must be `Dictionary<string, object?>`
- `Obsolete` methods and properties were removed
	- `MapView.OnMapRendered` - use `OnViewRendered`
	- `Widget.IconClass` (all widgets) - use `Icon`
	- `ExpandWidget.ExpandIconClass` - use `ExpandIcon`
	- `LocateWidget.UseHeadingEnabled` - use `RotateEnabled`
	- `ArealUnit` - use `GeometryEngineAreaUnit`
	- `GeometryEngine.GeodesicArea` with `ArealUnit` - use `GeodesicArea` with `GeometryEngineAreaUnit`
	- `GeometryEngine.PlanarArea` with `ArealUnit` - use `PlanarArea` with `GeometryEngineAreaUnit`
	- `LogicComponent.OnJavascriptErrorHandler` - calls to logic component methods will now throw exceptions just like any other .NET code
	- `AttributesDictionary` implicit conversion from `Dictionary<string, object>` - use the explicit conversion or constructor instead
- Namespace change for `ListItem`

# Performance Improvements
- Queries now return protobuf-serialized graphics in a stream
	- `FeatureLayer.QueryFeatures`
	- `FeatureLayer.QueryRelatedFeatures`,
	- `FeatureLayer.QueryTopFeatures`
	- `FeatureLayerView.QueryFeatures`
	- `MapView.HitTest`


test/dymaptic.GeoBlazor.Core.Test/dymaptic.GeoBlazor.Core.Test.csproj
```diff
@@ -19,7 +19,6 @@
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="SixLabors.ImageSharp" Version="3.0.2"/>
```
seahro, line 22: do we know where this came from, since it showed up in a number of repos?

TimPurdum: I added it to the unit tests when trying to implement a "snapshot comparison" Playwright test long ago. The test was always shaky and didn't work out.

It is used for handling attachments in some of our other apps.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/SymbolTests.razor
```diff
@@ -56,10 +57,10 @@
                                         VerticalAlignment="VerticalAlignment.Top"
                                         Kerning="false"
                                         LineHeight="2.3"
```
seahro, line 59: does this need to match the adjustment that was done with linewidth? or since this is neither a point nor pixel reference...it can stay as is

TimPurdum: Yeah, it's a weird one, but in CSS, `line-height` is not actually pixel based by default.

> The used value is this unitless [<number>](https://developer.mozilla.org/en-US/docs/Web/CSS/number) multiplied by the element's own font size. The computed value is the same as the specified <number>. In most cases, this is the preferred way to set line-height and avoid unexpected results due to inheritance.

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ManyGraphics.razor
```diff
@@ -21,8 +21,8 @@
         <SpatialReference Wkid="102100" />
     </Extent>
     <Map>
-        <OpenStreetMapLayer />
         <GraphicsLayer @ref="_graphicsLayer" />
+        <OpenStreetMapLayer />
```
seahro, line 25: I did not see anything here or in `Featurelayers` dont know if maybe it would be a todo or some information in the GB documentation to link to the ESRI docs, but an explanation of layer/data ordering to affect the display of those different items that would be in the operational layer.

src/dymaptic.GeoBlazor.Core/Components/Layers/OpacityVariable.cs
```diff
@@ -86,7 +116,7 @@ internal override void ValidateRequiredChildren()
         }
     }
 
-    private HashSet<OpacityStop>? _stops;
+    private List<OpacityStop>? _stops;
```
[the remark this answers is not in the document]
TimPurdum: Yes, the duplicates are unfortunate, but `HashSet` is not _ordered_, which means the stops (and many others I changed) would not stay in a given order after being set.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ManyGraphics.razor
TimPurdum: It is in the release notes, first item under `Breaking Changes`, but I agree, we should probably have a dedicated docs page.

src/dymaptic.GeoBlazor.Core/Components/Layers/OpacityVariable.cs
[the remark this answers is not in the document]
seahro: i didnt think about that , makes sense

src/dymaptic.GeoBlazor.Core/Components/Renderers/ImageryRenderer.cs
```diff
@@ -1,20 +1,60 @@
...
+            IDictionary<string, object?> temp)
+        {
+            return null;
+        }
+
+        if (temp.TryGetValue("type", out object? typeValue))
+        {
+            switch (typeValue?.ToString())
+            {
+                case "raster-stretch":
```
seahro, line 42: i will make notes of this implementation for these renderers, when looking to add more renderers.

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
```diff
@@ -2071,8 +2082,8 @@ protected override async Task RenderView(bool forceRender = false)
         }
 
         Rendering = true;
-        Map.Layers.RemoveWhere(l => l.Imported);
-        Map.Basemap?.Layers.RemoveWhere(l => l.Imported);
+        Map.Layers.RemoveAll(l => l.Imported);
```
seahro, line 2085: This required a lookup on my part and makes sense combined with the other transitions from the use of `Hashset<T>` (https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.removewhere?view=net-8.0) to `List<T>`(https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeall?view=net-8.0)

seahro (approved):
Looks good! Approved!

src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
TimPurdum: Yeah, this was just necessary due to the fact that they don't have the same remove methods.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ImageryTileLayers.razor
```diff
@@ -0,0 +1,163 @@
...
+    {
+        _stretchType = (StretchType)Enum.Parse(typeof(StretchType), e.Value?.ToString()!);
+
+        switch (_stretchType)
+        {
+            case StretchType.MinMax:
+                await _slider!.SetMax(5000);
+                await _slider.SetMin(0);
+                await _slider.SetStepInterval(10);
+                await _slider.SetValues([120, 3068]);
```
agnesstelmach, line 129: do these async methods need to finish in order? I was wondering if there was something that could be implemented that's similar to Promise.all() in JS, where an array of async promises can be fired at once and when all of them finish, continue on.

src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
```diff
@@ -77,9 +106,11 @@ public class BasemapLayerListWidget : Widget
     ///     Returns the modified base <see cref="ListItem" />
     /// </returns>
     [JSInvokable]
-    public Task<ListItem>? OnBaseListItemCreated(ListItem item)
+    public async Task<object> OnBaseListItemCreated(ListItem item)
```
agnesstelmach, line 109: just curious, why `ListItem` was changed to a more general `object`?

agnesstelmach (approved):

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/ImageryTileLayers.razor
TimPurdum: C# does have that feature, something like below, but it's wordy and I'm not sure if it's worth doing in this scenario.

```csharp
List<Task> tasks = [];
tasks.Add(Task.Run(() => _slider.SetMax(5000)));
tasks.Add(Task.Run(() => ...:
...
...

await Task.WhenAll(tasks);

src/dymaptic.GeoBlazor.Core/Components/Widgets/BasemapLayerListWidget.cs
TimPurdum: Good question, and I had to revisit to remember why. The return value of a `JSInvokable` is going back to JS, and we have issues sometimes with serialization when not casting to object first. You'll see a few other places in GeoBlazor where we cast to object before sending to JS. I will add a comment here.

PR #322, opened 2024-04-03
fix: layer JsModule is null after AddLayer

### LINKED ISSUE #321: Unable to call `SetVisibility` from a `GraphicsLayer` created in C#
If I create a `GraphicsLayer` in C# and add it to the `MapView`, I am unable to call `SetVisibility` and I assume other methods that require the `jsModule`.

Doing so, causes the following (tested on Blazor WebAssembly and MAUI Hybrid)
image
image

This is the code that fails
```c#
var secondPointsLayer = new GraphicsLayer();
await secondPointsLayer.Add(mapPoints);
await mapView.AddLayer(secondPointsLayer);

await secondPointsLayer.SetVisibility(false); // ERRORS
```

By adding the following line before `SetVisibility`, the code works as expected
```c#
var secondPointsLayer = new GraphicsLayer();
await secondPointsLayer.Add(mapPoints);
await mapView.AddLayer(secondPointsLayer);

secondPointsLayer.JsModule = mapView.JsModule; // ADD THIS LINE
await secondPointsLayer.SetVisibility(false); // WORKS AS EXPECTED
```

I'm happy to attempt to fix this, my first though is to check the module when `AddLayer` is called and if `null`, just set it - https://github.com/dymaptic/GeoBlazor/blob/1b8985a67ced91e3eb1a48138b1d8cebf346a207/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs#L1286-L1312

If you need a sample project, let me know.  Tested with v3.0.1

CyberDNS:
Hi dear GeoBlazor Team,

Description:
We noticed an issue while setting layer visibility.
This issue can be reproduced only if layer has been added by code to the MapView.
This issue does not occur if layer has been added using markup of the component (Razor language).

Changes:
To fix this issue, JsModule has been set while executing the AddLayer method.
New test method TestCanSetFeatureLayerVisibility has been added to validate the fix.

Impact:
This change fix the failure while setting layer visibility.

Resolves #321 

Br,
David from Team GIS - POST Luxembourg

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/Views/MapView.razor.cs
@@ -1295,6 +1295,7 @@ public partial class MapView : MapComponent
         {
             Map!.Layers.Add(layer);
             layer.Parent ??= Map;
+            layer.JsModule ??= ViewJsModule;
         }
 
         if (ViewJsModule is null) return;
--- a/test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
+++ b/test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
@@ -7,6 +7,32 @@
 @code {
 
     [TestMethod]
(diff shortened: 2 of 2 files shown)


TimPurdum (approved):
Looks great! Thank you again for your contribution!

PR #326, opened 2024-04-06
fix for MapFont serialization

### LINKED ISSUE #320: Add Many Graphics is doesn't work with TextSymbol with MapFont since Version 3.01
Add in ManyGraphics.razor following Code:

`    private Symbol GenerateText()
    {
        return new TextSymbol(
            "Hello",
            color: new MapColor("red"),
            backgroundColor: new MapColor("transparent")
             , font: new MapFont(12, "Arial", "normal", "bold")
            );
    }
`
In GenerateGraphic add:

`                case "Text":
                    return new Graphic(GeneratePoint(), GenerateText(), GeneratePopupTemplate(),
                        GenerateAttributes());
`
Replace this:

`    <select disabled="@_generating" @bind="_geometryType" @bind:after="SetGraphics">
        <option value="Point">Point</option>
        <option value="Polyline">Polyline</option>
        <option value="Polygon">Polygon</option>
        <option value="Text">Text</option>
    </select>
`

Run it

TimPurdum:
Closes #320 

MapFont serialization in Protobuf threw an exception because `Dimension` was not defined in the serialization.
Created a new `MapFontSerializationRecord` that converted Dimension to double, which matches what we do with other serializations using Dimension.


src/dymaptic.GeoBlazor.Core/Components/Symbols/MapFont.cs
```diff
@@ -38,7 +37,7 @@ public MapFont()
     /// <param name="weight">
     ///     The text weight.
     /// </param>
-    public MapFont(int? size, string? family, string? style, string? weight)
+    public MapFont(Dimension? size, string? family, string? style, string? weight)
```
seahro, line 40: `Dimension` or `double`? for size... looking at the other `size` references below

seahro (approved):
I did have a question about the MapFont using `Dimension` for size

src/dymaptic.GeoBlazor.Core/Components/Symbols/MapFont.cs
TimPurdum: `Dimension` for the C# class. We _serialize_ this in Protobuf as a `double`. Notice at the bottom of the file you are looking at `MapFontSerializationRecord` (protobuf representation).

PR #329, opened 2024-04-20
Demo/vs live

TimPurdum:
Fixes and improvements while working on Demo for VS Live Chicago

- Added implicit operator from `string` to `Effect`
- `GraphicsLayer` properties
   - `Effect`
   - `BlendMode`
   - `MinScale`
   - `MaxScale`
   - `ScreenSizePerspectiveEnabled`
- Moved `PersistenceEnabled` from `ImageryLayer`, `MapImageLayer`, and `ImageryTileLayer` to `Layer`
- `Sublayer`
   - `GetAllSublayers`
   - `SetProperty`
   - `SetLegendEnabled`
   - `SetPopupEnabled`
   - `SetPopupTemplate`
- Fix JsonConverter for `TileInfoFormat`
- `TileLayer`
   - `ApiKey`
   - `AllSublayers`
   - `Capabilities`
   - `Copyright`
   - `Effect`
   - `LegendEnabled`
   - `RefreshInterval`
   - `SourceJSON`
   - `SpatialReference`
   - `Sublayers`
   - `TileInfo`
   - `TileServers`
- Fix for JS callbacks after disposal in `MapView`
- Fix for invalid properties sent to `SetProperty`
- Fix for concurrency issue when disposing `AbortManager`
- Fix for `FeatureLayer` in 3D scenes

PR #330, opened 2024-04-21
Gh action testing

TimPurdum:
- Fix for build that was failing to include JS files in nuget packages.
- NOTE: I built this branch off of #329, the only file that needs to be reviewed separately here is `main-release-build.yml`

### DIFF
--- a/.github/workflows/main-release-build.yml
+++ b/.github/workflows/main-release-build.yml
@@ -46,9 +46,10 @@ jobs:
       # Verify that the JS files are present in the output directory
       run: |
         $nugetPath = "./src/dymaptic.GeoBlazor.Core/bin/Release/dymaptic.GeoBlazor.Core.${{env.CORE_VERSION}}.nupkg"
+        $extractPath = "./src/dymaptic.GeoBlazor.Core/bin/Release/extract"
         $jsFilePath = "staticwebassets\js\arcGisJsInterop.js"
-        Expand-Archive -Path $nugetPath -Force -DestinationPath $env:TEMP
-        $contents = Get-ChildItem -Path $env:TEMP -Recurse
+        Expand-Archive -Path $nugetPath -Force -DestinationPath $extractPath
+        $contents = Get-ChildItem -Path $extractPath -Recurse
         $fileExists = $contents -match [regex]::Escape($jsFilePath)
     - name: Build Templates
       run: dotnet build ./templates/dymaptic.GeoBlazor.Templates.csproj -c Release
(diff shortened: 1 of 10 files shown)


src/dymaptic.GeoBlazor.Core/Components/Layers/TileLayer.cs
```diff
@@ -16,13 +18,55 @@ public class TileLayer : Layer
     /// <inheritdoc />
     [JsonPropertyName("type")]
     public override string LayerType => "tile";
+    
+    /// <summary>
+    ///     An authorization string used to access a resource or service. API keys are generated and managed in the ArcGIS Developer dashboard. An API key is tied explicitly to an ArcGIS account; it is also used to monitor service usage. Setting a fine-grained API key on a specific class overrides the global API key.
+    /// </summary>
+    [Parameter]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? ApiKey { get; set; }
```
seahro, line 27: Do we want to move this to the layer class?  Then it can be inherited by other layer types as they are added in the future or only insert this for specific layers as needed, like this case

seahro (approved):
Looks good, couple questions

src/dymaptic.GeoBlazor.Core/Components/Layers/TileLayer.cs
TimPurdum: No, it's based on the inheritance pattern in ArcGIS. ApiKey is only available on some Layers.

src/dymaptic.GeoBlazor.Core/Objects/AbortManager.cs
```diff
@@ -69,32 +70,47 @@ public async Task<IJSObjectReference> CreateAbortSignal(CancellationToken cancel
...
-        if (_tokensAndControllers.ContainsKey(cancellationToken))
+            if (_tokensAndControllers.ContainsKey(cancellationToken))
+            {
+                _tokensAndControllers.TryRemove(cancellationToken, out _);
+            }
+        }
+        catch(Exception ex)
         {
-            _tokensAndControllers.Remove(cancellationToken);
+            Console.WriteLine(ex);
```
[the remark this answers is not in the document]
TimPurdum: Errors during object disposal usually have to do with race conditions of the object(s) already being disposed or set to `null`. In this case, I actually fixed the error I saw by switching to `TryRemove`, but I figured, even if we get an unknown error, we probably don't want to crash/throw. So in those cases, where I don't really expect anything, I tend to default to `Console.WriteLine`. I'm open to other suggestions, but if these are rare, a more detailed console message might not make a lot of difference, since it would be often ignored/missed.

PR #332, opened 2024-04-26
disposal and action button fixes

### LINKED ISSUE #331: Bug - MapView is not property disposed on navigation
Scenario
- Set up a MapView and Map with a GraphicsLayer
- Create a PopupTemplate with an ActionButton that calls back to a method in your razor file
- Navigate away from the page, then back
- Click the action button in the popup. You will see two firings, one for the old view, one for the new view.

TimPurdum:
Closes #331 

- Fixes for JS disposal of types
- Fixes for references to DotNet references in JS that were not being disposed correctly
- Improvements to Unit tests
- Unit test fixes

### DIFF
--- a/templates/dymaptic.GeoBlazor.Templates.csproj
+++ b/templates/dymaptic.GeoBlazor.Templates.csproj
@@ -3,7 +3,7 @@
 	<PropertyGroup>
 		<PackageId>dymaptic.GeoBlazor.Templates</PackageId>
 		<Title>GeoBlazor Project Templates</Title>
-		<PackageVersion>3.0.2-beta-2-beta-1</PackageVersion>
+		<PackageVersion>3.0.2-beta-11-beta-2-beta-1</PackageVersion>
 		<Version>2.5.2</Version>
 		<PackageLicenseExpression>MIT</PackageLicenseExpression>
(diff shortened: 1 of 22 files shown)


templates/dymaptic.GeoBlazor.Templates.csproj
```diff
@@ -3,7 +3,7 @@
 	<PropertyGroup>
 		<PackageId>dymaptic.GeoBlazor.Templates</PackageId>
 		<Title>GeoBlazor Project Templates</Title>
-		<PackageVersion>3.0.2-beta-2-beta-1</PackageVersion>
+		<PackageVersion>3.0.2-beta-11-beta-2-beta-1</PackageVersion>
```
AndersenBell, line 6: what do the -beta2-beta1 mean here?

AndersenBell (approved):

zwaap (approved):
Looks good as far as I can see

templates/dymaptic.GeoBlazor.Templates.csproj
TimPurdum: I have a powershell script that is updating the version in the templates, looks like it is broken. Good catch!

PR #340, opened 2024-05-16
Adjust Default Inset Values for Better Alignment of CustomOverlay

CyberDNS:
This pull request adjusts the default inset values in the CustomOverlay component to improve alignment with standard ArcGIS UI widgets.

The original inset values "inset: 30px 15px;" caused the overlays to be positioned slightly lower than expected, misaligning them with elements like the ArcGIS zoom widget at the top of the screen. The corrected inset is now "inset: 15px 15px 30px;”

image

Br,
David for Team GIS Post Luxembourg


TimPurdum:
@CyberDNS looks pretty straightforward. I do want to pull this and check against the samples where we have our dymaptic/GeoBlazor inset in that top right, to avoid overlap there. It looks like you're hiding that, which is totally fine. Will get back soon.

TimPurdum:
My concern was unwarranted, there's plenty of space for the branding.

TimPurdum (approved):

PR #349, opened 2024-07-21
Bug/319 localization double parse

### LINKED ISSUE #319: Localization Issue with comma decimal and serialization
From Discord report:

> Hello. I think you have a localisation issue. I Noway we use ', ' comma as decimal separator and data is stored in database with comma.
and thus the program crash i the QueryFeatures call as shown below. If I change the system settings to use pint as decimal separator the QueryFeatures is OK.

image

I believe `double.Parse` can take a `CultureInfo` to determine how to parse. However, I wonder if knowing _which_ culture to use might be an issue because we need to know the culture used in generating the graphic attribute data in the first place, which could come from various sources.

TimPurdum:
Closes #319 
- Passes `navigator.language` with graphic attributes in `dotNetBuilder.buildDotNetGraphic`
- Uses the passed culture string if available to parse numbers and datetimes. Falls back to `CurrentCulture`.
- Added new unit tests to validate parsing
- Added support in `TestRunnerBase.razor` for synchronous tests

There is some question in my mind still if this is the "right" or "best" approach. The issue IMO stems from the following:

- Data sources, such as AGOL, can store feature data with a culture
- Client devices can have a different culture
- GeoBlazor servers could in theory have yet another culture

How these 3 pieces fit together, and how to _know_ that we aren't breaking data in any one scenario (data from AGOL in culture A, data from GeoBlazor in culture B, etc.).

For example, if we pass data to ArcGIS JS layers in culture A _from_ GeoBlazor, but the client is in culture B, when it returns to GB, it will have culture B tied to the data. Some of this stems from the fact that Blazor is both a server and client technology, often in the same application.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Objects/AttributesDictionary.cs
+++ b/src/dymaptic.GeoBlazor.Core/Objects/AttributesDictionary.cs
@@ -1,9 +1,6 @@
 ﻿using dymaptic.GeoBlazor.Core.Components;
 using ProtoBuf;
-using System.Diagnostics;
 using System.Globalization;
-using System.Reflection;
-using System.Reflection.Metadata;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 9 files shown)


TimPurdum:
Questions above aside, these changes will make GeoBlazor work inline with the ArcGIS JS SDK, along with a few other options that you could use to manipulate the `AttributesDictionary` server-side. I think that is a step in the right direction, it fixes the issue raised by an end user in Europe, and it should be merged.

src/dymaptic.GeoBlazor.Core/Objects/AttributesDictionary.cs
```diff
@@ -16,21 +17,28 @@ namespace dymaptic.GeoBlazor.Core.Objects;
...
-    ///     The dictionary to use
+    ///     The dictionary to use.
     /// </param>
+    /// <remarks>
+    ///     It is possible to control the culture used for parsing numbers and dates by including a "geoBlazorCulture" key in the dictionary. The value should be a string representation of a culture name, such as "en-US" or "fr-FR".
+    ///     Alternatively, for a server-generated dictionary, the culture can be set in the server's CultureInfo.DefaultThreadCurrentUICulture property. It will fall back to `CultureInfo.CurrentCulture` if not set.
+    /// </remarks>
     public AttributesDictionary(Dictionary<string, object?> dictionary)
     {
+        CultureInfo cultureInfo = dictionary.TryGetValue("geoBlazorCulture", out object? gbCulture)
```
AndersenBell, line 39: Would it make sense to make this a property on the class rather then another entry in the dictionary? It could help separate data vs framework, unless these attributes are not directly realted to the data.

AndersenBell (approved):

src/dymaptic.GeoBlazor.Core/Objects/AttributesDictionary.cs
TimPurdum: The problem is the serialization of the attributes, where we have to switch them to an array of objects with `key`, `value`, and `valueType`. I could rewrite all of that to be an object _with_ an array, but this seemed simpler. Rewriting would mean rewriting both JSON and Protobuf serialization methods in both C# and TypeScript. 😩

TimPurdum:
After discussing with @AndersenBell, we came up with a better plan. In my changes, we now query once for the browser language in `JsModuleManager` when the js modules are loaded. This is then referenced in `AttributesDictionary` to deserialize, and we no longer need to add the culture to each attribute dictionary.

AndersenBell (approved):
looks good

PR #365, opened 2024-09-26
Feature/59 area measurement 2d

seahro:
Adds the 2d Area Measurement class and area measurement widget to the widget demo page.


agnesstelmach (approved):
looks good, though I seeing warnings from github here talking about missing XML comments.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -2155,6 +2157,14 @@ async function createWidget(widget: any, viewId: string): Promise<Widget | null>
                 icon: widget.icon ?? undefined,
             });
             break;
+        case 'areaMeasurement2D':
+            newWidget = new AreaMeasurement2D({
+                view: view,
+                viewModel: widget.AreaMeasurement2DViewModel ?? undefined,
+                unit: widget.unit ?? undefined,
+                unitOptions: widget.unitOptions ?? undefined
+            });
```
TimPurdum, line 2166: Can you add unit tests that test with/without the unit and options? I realize we are probably waiting on the ViewModel as we have for other types.

TimPurdum (changes requested):

TimPurdum:
> looks good, though I seeing warnings from github here talking about missing XML comments.

Yeah, these are required but the warning as error only shows on release mode, so they're easy to miss. Please add XML comments to all public members and types.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
seahro: in an earlier commit, i did get start building out the viewmodel.  but decided against it going through it, since we have those other ouststanding viewmodels to do also.  I think once I complete this PR.  I think we can add this ViewModel in another PR and start to knock out those other 4... so we can start some to add some customization

seahro:
I had warnings as errors turned off while troubleshooting building the Core project and getting nuget vulnerability warnings related to Markdig dependencies.  XML comments have been added in. 
-3 new unit tests for the measurement widget are now in place

TimPurdum (approved):

TimPurdum (approved):

PR #369, opened 2024-12-11
Updating SpatialReference class

AndersenBell:
Spatial References support setting WKID or WKT or WKT2.

Fixed a bug in the transition between .net and Javascript with a SpatialReference with only WKT.

Fixed a bug in the in the transition between .net and Javascript with a SpatialReference with a null WKID

Added support WKT2 in GeoBlazor

Added Tests for different SpatialReference types

Updated samples to no longer use a deprecated property.

### DIFF
--- a/test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/SpatialReferenceTests.razor
+++ b/test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/SpatialReferenceTests.razor
@@ -1,17 +1,58 @@
 ﻿@inherits TestRunnerBase
 
+@inject Projection projection
+
 @{
     base.BuildRenderTree(__builder);
 }
 
 @code {
-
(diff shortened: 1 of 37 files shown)


test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/SpatialReferenceTests.razor
```diff
@@ -0,0 +1,58 @@
+﻿@inherits TestRunnerBase
+
+@inject Projection projection
+
+@{
+    base.BuildRenderTree(__builder);
+}
```
TimPurdum, line 7: You can move this code into a .cs file and remove the `renderHandler` parameters since you're not instantiating a mapview. See `GeometryEngineTests.cs` for example.

TimPurdum (approved):
Looks good! One suggestion.

PR #376, opened 2025-01-31
Release v3.2.1

TimPurdum:
# Release Notes

## Bug Fix
- Fixes `AttributesDictionary` value deserialization when querying, using hit tests, or sketch widget


TimPurdum:
The widget typescript change is mostly irrelevant, just trying to clean up the resulting HTML. I have tested that change in our client application as well to verify it does not break.

PR #378, opened 2025-02-03
Updating JS Point and Bookmark creation

### LINKED ISSUE #377: Bug: Bookmark widget bookmarks make the map crash
On the Geoblazor Samples page go to widgets, select bookmarks and then click on a bookmark. The map will disappear. If you click the refresh button, it will reload the map at the bookmark location correctly.

There are no errors in the console

This is also found in 3.1.0

AndersenBell:
closes #377

A resent update to the ArcGIS libraries changed the way that `rotation` is handled on `ViewPoints`.
Updating the buildJsViewpoint to use the newer `copyValuesIfExists` addresses this.

`buildJsPoint` was also updated to use `copyValuesIfExists`

A few other minor cleanup things too.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Licensing.cs
+++ b/src/dymaptic.GeoBlazor.Core/Licensing.cs
@@ -9,7 +9,7 @@ internal static class Licensing
     {
         try
         {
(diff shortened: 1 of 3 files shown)


TimPurdum (approved):
Looks good!

PR #379, opened 2025-02-06
Feature/update jsbuilder

AndersenBell:
This PR is a follow up to https://github.com/dymaptic/GeoBlazor/pull/378
Which found a bug in the way default properties were set for some esri objects.
This PR attempts to fix most of that by migrating as much as possible of the js initialization to use the `copyValuesIfExists` function to only change the default property value if there is one set on the dotNet object.

This should hopefully prevent future occurrences of similar issues where default values change on the JavaScript side and introduce previously non-existent bugs.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
+++ b/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
@@ -129,7 +129,7 @@ import FeatureLayerViewWrapper from "./featureLayerView";
 import Popup from "@arcgis/core/widgets/Popup";
 import ElevationLayer from "@arcgis/core/layers/ElevationLayer";
 import PopupWidgetWrapper from "./popupWidgetWrapper";
-import {load} from "protobufjs";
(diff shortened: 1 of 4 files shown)


src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -3024,11 +3024,13 @@ function buildDotNetListItem(item: ListItem): DotNetListItem | null {
 
 // this function should only be used for simple types that are guaranteed to succeed in serialization and conversion
 export function copyValuesIfExists(originalObject: any, newObject: any, ...params: Array<string>) {
-    params.forEach(p => {
-        if (hasValue(originalObject[p]) && originalObject[p] !== newObject[p]) {
-            newObject[p] = originalObject[p];
-        }
-    });
+    if (hasValue(originalObject)) {
```
TimPurdum, line 3027: good catch, this is a useful check!

TimPurdum (approved):

PR #381, opened 2025-02-12
Change Content to None to exclude files.

### LINKED ISSUE #380: Remove package.json and tsconfig.json from nuget packages
Requested on Discord, a user reported that the package.json and tsconfig.json files in the packages were causing issues. These shouldn't be needed since we are building JS before we pack.

TimPurdum:
Closes #380


AndersenBell (approved):
looks good

PR #385, opened 2025-03-27
Fix OpenStreetMapLayer urlTemplate Error with WebTileLayer

Pouria7:
#### Issue
The ArcGIS JS API core throws a `"Cannot assign to read-only property 'urlTemplate'"` error when setting a custom `urlTemplate` on `OpenStreetMapLayer`, even in the constructor. This appears to be an API limitation.

#### Fix
This PR switches to `WebTileLayer` for the `open-street-map` case when a custom `urlTemplate` is provided, then copies relevant properties (`subDomains`, `blendMode`, `copyright`, `maxScale`, `minScale`, `fullExtent`) from an `OpenStreetMapLayer` instance to maintain consistent behavior. While the root issue lies in the ArcGIS JS API, this approach effectively resolves it for the wrapper.

#### Notes
Tested and works with custom tile sources (e.g., offline proxies). No functionality is lost.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
+++ b/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
@@ -2674,11 +2674,14 @@ export async function createLayer(layerObject: any, wrap?: boolean | null, viewI
 
             break;
         case 'open-street-map':
-            let openStreetMapLayer: OpenStreetMapLayer;
+            let openStreetMapLayer: WebTileLayer;
             if (hasValue(layerObject.urlTemplate)) {
-                openStreetMapLayer = new OpenStreetMapLayer({
+                openStreetMapLayer = new WebTileLayer({
                     urlTemplate: layerObject.urlTemplate
                 });
+                copyValuesIfExists(new OpenStreetMapLayer(), openStreetMapLayer,
+                    'subDomains', 'blendMode', 'copyright', 'maxScale', 'minScale', 'fullExtent'
+                )
             } else if (hasValue(layerObject.portalItem)) {
                 let portalItem = buildJsPortalItem(layerObject.portalItem);
                 openStreetMapLayer = new OpenStreetMapLayer({ portalItem: portalItem });


TimPurdum:
@Pouria7 this looks good, nice workaround. I'm just a few days away from merging in #384, and this logic would need to be moved to `openStreetMapLayer.ts`. We will now have auto-generated files from ArcGIS (.gb.cs & .gb.ts), and you can override the functionality by changing the .ts file that calls the generated file. If you have time to look at adding your changes there, to ahead. Otherwise I can add it here shortly.

TimPurdum:
@Pouria7 is there a source we could add to a unit test for this functionality?

Pouria7:
@TimPurdum Sounds good! I’ll update the PR to move this logic to openStreetMapLayer.ts once #384 is merged. Let me know if there are any specific considerations I should keep in mind while making the changes.
I don’t have an existing unit test for this, but I do have a minimal reproducible project that demonstrates the issue.
[Bug-Repo](https://github.com/Pouria7/GeoBlazor-OpenStreetMap-Bug-Repro)
Let me know if this helps or if you’d like me to integrate a test case directly

TimPurdum:
@Pouria7 #384 is merged. I'm still working on documentation for the new pattern, but here's the gist.

Most classes in C# and TS will have two files. The one with the normal extension (.cs, .ts) is "human editable", and the one with .gb.cs or .gb.ts is generated code.

For TypeScript overriding, you would do something like this in `openStreetMapLayer.ts`

take this method:

```ts
export async function buildJsOpenStreetMapLayer(dotNetObject: any, layerId: string | null, viewId: string | null): Promise<any> {
    let { buildJsOpenStreetMapLayerGenerated } = await import('./openStreetMapLayer.gb');
    return await buildJsOpenStreetMapLayerGenerated(dotNetObject, layerId, viewId);
}
```

and inject your code before the call to the generated file. You can import and call `buildJsWebTileLayer` from here with the same pattern (call `buildJsWebTileLayer` not `buildJsWebTileLayerGenerated` so that we can make future override edits there).

If you can, please do add a test as well. This would go in `\test\dymaptic.GeoBlazor.Core.Test.Blazor.Shared\Components\LayerTests.razor`. Hopefully the pattern there should be fairly clear on how to assert that the layer is rendered properly. You can use the same markup as in your repro project.

TimPurdum:
Oh, I forgot. I presume you're already compiling the TypeScript for testing purposes. You might need to run `npm install` once to get updated packages.

TimPurdum:
We had a few other buggy things in develop that I fixed this afternoon, FYI if you're hitting any roadblocks merge in again from develop.

TimPurdum:
Closing this in favor of #390 which is the newer version.

PR #387, opened 2025-04-01
V4 documentation

TimPurdum:
A few fixes found during generating release notes.

PR #328, opened 2024-04-15
Improving Map Refresh

CyberDNS:
Hi,

We made a change that lets you update maps without reloading everything. This helps keep the map working smoothly when you change it.

Resolves #327 

Br,
David for GIS Team Post Luxembourg

Issue #327, opened 2024-04-15
Enhance Basemap Refresh Functionality


CyberDNS:
We have identified a need for improved basemap refresh capabilities. 
Currently, updating basemaps portalItem or styles requires a full mapView refresh, like you do it in the sample page basemaps, 
which isn't ideal for preserving the user's application state (like measurements or annotations...).
 
We propose an update that supports more specific basemap refreshes, allowing for seamless changes without a full mapView reload. This enhancement would significantly benefit applications requiring dynamic map updates while maintaining the user context. 
We plan to submit a pull request with a proposed implementation for this feature.

Br,
David for GIS Team POST Luxembourg

PR #388, opened 2025-04-01
Feature/327 map refresh

TimPurdum:
*Please review and merge #387, then update this branch from `develop`. It includes the changes from that PR as well, and this will cut down on the number of changes.*

Closing #327, but not necessarily fully implementing as requested in #328.

The original attempt there was to update the Blazor Parameter for the PortalItem Id. However, as we have discovered more and more over time, this is challenging to do well and keep in sync. Instead, I fully implemented and tested the `portalItem.SetPortalItemId()` method and updated the tests to match.

While doing this, I ran across and tackled a few other issues:

- `WebMap`/`WebScene`
  - PortalItem is valid in many layers, as well as WebMap and WebScene
  - Added the `IPortalLayer` interface to WebMap and WebScene
  - Copied the implementations for `PortalItem`, set/get methods, and registration into WebMap/WebScene
  - Removed WebMap/WebScene from the code generation exclusion list, and generated code for each. This generated a bunch of new types
  - Updated the startup TypeScript in `arcGisJsInterop.buildMapView()` to call the generated code
  - Re-tested all tests and samples that use WebMap and WebScene
- `SetProperty`/`GetProperty`
  - In generated code, when there is no dedicated TypeScript wrapper and method, we call `setProperty` and `getProperty`. These were incorrectly using the GeoBlazor version of the property name, not the ArcGIS version. Updated this and re-ran code generation. This made a lot of single-line changes in `.gb.cs` files.


src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.gb.cs
```diff
@@ -1438,17 +1438,17 @@ public async Task OnJsPixelFilter(PixelData pixelData)
             return MultidimensionalInfo;
         }
 
-        // get the property value
-        RasterMultidimensionalInfo? result = await JsComponentReference!.InvokeAsync<RasterMultidimensionalInfo?>("getProperty",
-            CancellationTokenSource.Token, "multidimensionalInfo");
+        RasterMultidimensionalInfo? result = await JsComponentReference.InvokeAsync<RasterMultidimensionalInfo?>(
+            "getMultidimensionalInfo", CancellationTokenSource.Token);
```
AndersenBell, line 1442: Is there a reason the `getMultidimensionalInfo` and `CancellationToken` parameters got switched? It looks like most other calls have them flipped?

src/dymaptic.GeoBlazor.Core/Components/Widgets/SearchWidget.gb.cs
```diff
@@ -269,7 +269,7 @@ public SearchWidget(
     [ArcGISProperty]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     [JsonInclude]
-    public IReadOnlyList<object>? Results { get; protected set; }
+    public SearchResultResponse? Results { get; protected set; }
```
AndersenBell, line 272: does the property description need updated too?

src/dymaptic.GeoBlazor.Core/Components/KMLLayerViewMapImage.gb.cs
```diff
@@ -303,7 +303,7 @@ public async Task SetExtent(Extent? value)
         }
         
         await CoreJsModule.InvokeVoidAsync("setProperty", CancellationTokenSource.Token,
-            JsComponentReference, "extent", value);
+            JsComponentReference, "Extent", value);
```
AndersenBell, line 306: why was this capitalized, most of the properties are lowercase?

src/dymaptic.GeoBlazor.Core/Components/WFSOperations.gb.cs
```diff
@@ -280,7 +280,7 @@ public async Task SetGetCapabilities(WFSOperationsGetCapabilities? value)
         }
         
         await CoreJsModule.InvokeVoidAsync("setProperty", CancellationTokenSource.Token,
-            JsComponentReference, "getCapabilities", value);
+            JsComponentReference, "GetCapabilities", value);
```
AndersenBell, line 283: why are all these capitalized?

AndersenBell (approved):

AndersenBell (approved):
Just a couple questions on some of the generated files

src/dymaptic.GeoBlazor.Core/Components/Layers/ImageryLayer.gb.cs
TimPurdum: If you look carefully, there's 2 parameters now instead of 3. The pattern for `InvokeAsync` is (`methodName`, `CancellationToken`, `parameters`).

There are two ways we pull properties. One is with the generic `getProperty` that takes a parameter. The other is with the explicit-typed methods like `getMultidimensionalInfo` that don't need a property. This is based on whether the type can be passed back directly from ArcGIS or it needs to call a builder in between.

src/dymaptic.GeoBlazor.Core/Components/Widgets/SearchWidget.gb.cs
TimPurdum: Hmm, yeah, it probably should. I don't have a rule for overriding descriptions in the code-generator, I'd have to move the entire property to the other file, or add such a rule.

src/dymaptic.GeoBlazor.Core/Components/KMLLayerViewMapImage.gb.cs
TimPurdum: image
Doesn't seem _correct_, but that is a capitalized property in the typescript definitions. I'm digging for a way to identify which is correct, but it's hard in the obfuscated JS source.

src/dymaptic.GeoBlazor.Core/Components/WFSOperations.gb.cs
TimPurdum: again, it's coming straight from the TS definitions. Frustrating that they are irregular or incorrect.

TimPurdum: These ones match in the online docs. The Extent one is lowercase in the docs. 😐

TimPurdum: https://github.com/dymaptic/GeoBlazor/issues/392

PR #389, opened 2025-04-02
removed generated files without root

seahro:
There were a number of generated files in the Geoblazor repos which were missing root files.  This removes those identified generated files.
`SlideGround.gb.cs` removed


TimPurdum:
@seahro this actually came through with only one file removed. Is that correct?

TimPurdum (approved):

seahro:
@TimPurdum yes, core only had the one "generated" file without a root.  Pro had a bunch of them

PR #383, opened 2025-02-13
Bookmark widget updates

AndersenBell:
closes #362 

Removes the rubber-banding effect when using widgets and also binding the lat/long on the MapView

The bookmark widget has deprecated `editingEnabled` this PR marks the same property in Geoblazor. 
It also adds a new class that replaces the `editingEnabled` property.
Fixing and updating the samples to use the new class

PR #393, opened 2025-04-11
Bug/362 bookmarks widgets v4

### LINKED ISSUE #362: LocateWidget resets on bound parameters of MapView
Similar to the #276, any widget like `LocateWidget` that redefines the map view Extent should also disable extent-defining parameters in `MapView`. Currently, if you bind the mapview lat/long and click the Locate widget, it will snap back to the defined lat/long.

TimPurdum:
Closes #362, follow up to #383 

Relevant Changes
1. The "delayed registration" logic in Layer and Widget classes in `RegisterChildComponent` has been updated to now call the relevant `Update` method immediately, but then check and see if it should delay the call to the JS update method.
2. In `MapView.AddWidget` and `MapView.AddLayer` we now also wait for all child components to register before calling to add the layer or widget to the map. (this is for widgets/layers that are added after the first map load). Note that I intentionally left `AddLayer` and `AddWidget` as `Task` methods, even though they don't actually need that now, to be consistent with other methods, and to not break the API in case we need to update the logic to an async call again in the future.

Unrelated Changes for issues found during testing
3. Added some missing checks for `CoreJsModule` and `JsComponentReference` in generated code methods.
4. Added logic to instantiate `CoreJsModule` inside `record` types that have methods and require it. Relevant change is in `MapComponent.OnParametersSet` and the new interface `IInteractiveRecord`.
5. Fixed a broken merge in `Core.csproj`


TimPurdum:
Need to go ahead and merge this since `develop` is currently broken. We can continue testing on develop.

PR #398, opened 2025-04-30
Bug/397 font size

### LINKED ISSUE #397: Customer Testing Bug Fixes
Two issues found with our current customer:

1. Updating the Font Size of a TextSymbol with `await graphic.SetSymbol` does not always work correctly. Issues is the naming of the `font` property in the protobuf serialization (it is named MapFont, but should be Font).
2. Ran into a race condition where `MapRendered` is false, meaning we have not yet returned from `MapView.RenderView`, yet `OnViewRendered` has been called. The callback should wait for the render method to complete before proceeding so that everything is aligned.

TimPurdum:
Closes #397 

- Changed `SymbolSerializationRecord.MapFont` name to `SymbolSerializationRecord.Font` to match expected ArcGIS property name when serializing via Protobuf. Added a unit test in `SymbolTests` to validate.
- In `MapView.OnJsViewRendered`, added a while loop to wait for `RenderView` to complete and `MapRendered` to be set. This prevents race conditions where the boolean flag is not set until after the callback is called. Added a unit test in `MapViewTests` to validate.
- Added `Id` to most Protobuf serializations, since this helps us match round-tripped objects (discovered when writing new TextSymbol font test). Exceptions: `SpatialReference`, since it uses wkid/wkt equality check, `MapColor`, `MapPath`, `Attribute` and record types with no `Id`. Also updated `graphic.json` with these Ids (this file is used to read the protobuf schema in the TypeScript layer.
  - `GeometrySerializationRecord`
  - `ActionBaseSerializationRecord`
  - `PopupContentSerializationRecord`
  - `PopupExpressionInfoSerializationRecord` (serialized as `ExpressionInfo`)
  - `SymbolSerializationRecord`
  - `ChartMediaInfoValueSerializationRecord`
  - `ChartMediaInfoValueSeriesSerializationRecord`
  - `FieldInfoSerializationRecord`
  - `FieldInfoFormatSerializationRecord`
  - `MediaInfoSerializationRecord`
  - `RelatedRecordsInfoFieldOrderSerializationRecord`


src/dymaptic.GeoBlazor.Core/Components/ActionBase.cs
```diff
@@ -75,7 +75,7 @@ public ActionBaseSerializationRecord()
     {
     }
     
-    public ActionBaseSerializationRecord(Guid Id,
+    public ActionBaseSerializationRecord(string Id,
```
AndersenBell, line 78: there seems to be inconsistencies between using a string or Guid for Id, is there a better way or a plan to normalize it?

AndersenBell (approved):

PR #400, opened 2025-05-01
update sublayer to copy nested sublayers

### LINKED ISSUE #399: Error in Sublayer.UpdateFromJavaScript
Reported in Discord, transcript below. The user is getting an error when a second nested "sublayer" is invoked from a parent sublayer.

Looking through the code, I don't immediately see the issue, but I did find some code to clean up. WIll also add unit tests and try to recreate the issue.

city of eugene
OP
 — Yesterday at 1:25 PM
I am getting a javascript error when I try to load a tilelayer from an on-prem arcgis server.  The tile layer loads and seems to work as expected so I am not sure why this layer is throwing an error.  Any suggestions or direction you could provide to get rid of this error is appreciated.  This tilemap was made using arcgis pro.

the razor page is 

    ```html
    <MapView Class="h-100 w-100" Zoom="2">
         <Map>
             <TileLayer Url="[http://server/arcgis/rest/services/folder/Basemap/MapServer"/>]
         </Map>
     </MapView>
    ```

and the error is 

```
Uncaught (in promise) Error: System.NullReferenceException: Object reference not set to an instance of an object.
   at dymaptic.GeoBlazor.Core.Components.Layers.Sublayer.UpdateFromJavaScript(Sublayer renderedSublayer)
   at dymaptic.GeoBlazor.Core.Components.Layers.Sublayer.UpdateFromJavaScript(Sublayer renderedSublayer)
   at dymaptic.GeoBlazor.Core.Components.Layers.TileLayer.UpdateFromJavaScript(Layer renderedLayer)
   at dymaptic.GeoBlazor.Core.Components.Views.MapView.OnJavascriptLayerViewCreate(LayerViewCreateInternalEvent layerViewCreateEvent)
    at b.endInvokeDotNetFromJS (blazor.webassembly.js:1:3136)
    at Object.gn [as endInvokeDotNetFromJS] (blazor.webassembly.js:1:58943)
    at invoke-js.ts:233:31
    at Ll (invoke-js.ts:276:5)
    at 00b21cfe:0x1f99d
    at 00b21cfe:0x1be44
    at 00b21cfe:0xf011
    at 00b21cfe:0x1e693
    at 00b21cfe:0x1ee89
    at 00b21cfe:0xce61
```

Tim, Chief BlazorBuilder — Yesterday at 1:33 PM
Are you using GeoBlazor 3.2.1, or the 4.0.0 beta?

city of eugene
OP
 — Yesterday at 2:39 PM
3.2.1

Tim, Chief BlazorBuilder — Yesterday at 3:11 PM
OK, I'm not sure exactly what is erroring there. Basically it's trying to read your sublayers and copy the properties to GeoBlazor so you could reference them if needed. It looks like it's happening when we have nested sublayers. I would suggest trying the 4.0 beta, although be aware there are numerous breaking changes. I'll post the release notes here in a minute.

If that doesn't work for you, or causes other issues, you should be able to ignore the JS error, if it's not actually breaking the site. I realize that's not ideal. I'd also love any details on your layer that you can share about how it is structured with sublayers, so I might be able to recreate and test further.

city of eugene
OP
 — 9:11 AM
here is how the layers are organized

Image

city of eugene
OP
 — 10:57 AM
I installed beta 4.0 and am seeing the same exception but there is a reference to some js code now.

Image

clicking the link taks me to     

```js
e.on("layerview-create", async s => {
        let o = Object.keys(z).find(y => z[y] === s.layer)
          , n = !1;
        M(e.map.basemap) && (e.map.basemap.baseLayers?.includes(s.layer) || e.map.basemap.referenceLayers?.includes(s.layer)) && (n = !0);
        let a, c;
        a = DotNet.createJSObjectReference(xb(s.layer)),
        c = DotNet.createJSObjectReference(xb(s.layerView));
        let p = {
            layerObjectRef: a,
            layerViewObjectRef: c,
            layerView: V8(s.layerView),
            layer: vE(s.layer, !1),
            layerGeoBlazorId: o,
            isBasemapLayer: n
        }
          , h = s.layer.id;
        if (Gy.includes(h))
            return;
        if (Gy.push(h),
        !T9) {
            await t.invokeMethodAsync("OnJavascriptLayerViewCreate", p),
            Gy.splice(Gy.indexOf(h), 1);
            return
        }
        let m = JSON.stringify(p.layer)
          , d = JSON.stringify(p.layerView)
          , f = 1e3
          , g = Math.ceil(m.length / f);
        for (let y = 0; y < g; y++) {
            let  = m.slice(y * f, (y + 1) * f);
            await t.invokeMethodAsync("OnJavascriptLayerCreateChunk", h, , y)
        }
        g = Math.ceil(d.length / f);
        for (let y = 0; y < g; y++) {
            let  = d.slice(y * f, (y + 1) * f);
            await t.invokeMethodAsync("OnJavascriptLayerViewCreateChunk", h, , y)
        }
        await t.invokeMethodAsync("OnJavascriptLayerViewCreateComplete", o ?? null, h, p.layerObjectRef, p.layerViewObjectRef, n),
        Gy.splice(Gy.indexOf(h), 1)
    }
```

Image

TimPurdum:
Closes #399 

Upon testing, discovered that nested sublayers were not being copied in the TS layer in the layerview-create callback, which _might_ have caused the user issue, although I could not recreate that.

- Updated `sublayer.ts` to copy nested sublayers
- Updated `TileLayer.cs` to remove incorrect direct clone of the `Sublayers`, which are walked through below that
- Added a new `TileLayer` and `MapImageLayer` test, each with nested sublayers
- Split apart `LayerTests` into separate layer types
- Updated a nuget package for build conflict with pro packages

The original error was a `NullReferenceException` in `Sublayer.UpdateFromJavaScript`. I inspected that, and the only way I can see that a type would be null is if there was a null inside the collection of sublayers. But I was not able to recreate this, before or after my fixes.

There are variables that _look_ like they could be null, such as `CoreJsModule`, but before calling `UpdateFromJavaScript`, we always set `Parent`, and the child looks up the module from the parent.


src/dymaptic.GeoBlazor.Core/Components/Layers/TileLayer.cs
```diff
@@ -80,7 +80,6 @@ internal override async Task UpdateFromJavaScript(Layer renderedLayer)
     {
         await base.UpdateFromJavaScript(renderedLayer);
         TileLayer renderedTileLayer = (TileLayer)renderedLayer;
-        Sublayers ??= renderedTileLayer.Sublayers;
```
AndersenBell, line 83: maybe add a comment to the Sublayers below why or what its doing

PR #403, opened 2025-05-15
update project files and readme

### LINKED ISSUE #402: Add CompressionEnabled False to all projects in Debug mode
During testing this week, I discovered that the following line, which we had tested some in the past, makes a huge difference in build times, so I think we should add it to all sample and test projects, as well as to our getting started documentation.

```xml
    <PropertyGroup>
        <CompressionEnabled Condition="$(Configuration) != 'RELEASE'">false</CompressionEnabled>
    </PropertyGroup>
```

TimPurdum:
Closes #402 

- Adds `CompressionEnabled = false` to `Debug` mode for all test and sample projects and libraries. This speeds up build times, which improves the development cycle. Some of them had the inverse setting (`Debug == true` instead of `Release != true`), so I updated these to be consistent.
- Added the `CompressionEnabled` setting as a tip to the two `ReadMe.md` files.
- In the internal `src/dymaptic.GeoBlazor.Core/ReadMe.md` file, I removed the html header material, because it was not being rendered correctly on NuGet.org when packaged. I left this material in the root `ReadMe.md` for GitHub.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <CoreVersion>4.0.0.0-beta-8</CoreVersion>
+    <CoreVersion>4.0.0.0-beta-9</CoreVersion>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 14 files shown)


codergrl (approved):
👍🏻

PR #407, opened 2025-05-19
Update readme

codergrl:
Updating the readme for the develop branch to let users know this applies to version 4. Also update code example to include Interactive rendering info


ReadMe.md
```diff
@@ -47,6 +47,9 @@ GeoBlazor brings the power of the ArcGIS Maps SDK for JavaScript into your Blazo
 
 [![Discord](https://img.shields.io/discord/1027907220949717033?color=%235865F2&label=chat&logo=discord&logoColor=white)](https://discord.gg/hcmbPzn4VW)
 
+## ‼️Important Version Information
+The below documentation is for GeoBlazor version 4.x. To view the documentation for previous versions, switch to the main branch, or go to the [Documentation](https://docs.geoblazor.com) page.
+
```
TimPurdum, line 52: This will be out of date as soon as we release v4, because we will merge into main. I suggest we point them to https://www.nuget.org/packages/dymaptic.GeoBlazor.Core/3.2.1#readme-body-tab, which will remain permanently locked to v3.

ReadMe.md
```diff
@@ -157,11 +160,18 @@ Or for the Pro version with additional features:
    ```csharp
    builder.Services.AddGeoBlazor(builder.Configuration);
    ```
+8. GeoBlazor requires Interactive server-side rendering enabled. Verify that the following lines are present in your `Program.cs`:
+
+   ```csharp
+      .AddInteractiveServerComponents(); // this should be in the Services section
 
+      .AddInteractiveServerRenderMode(); // this should be in the app section
+   ```
```
TimPurdum, line 169: I realize they are stacked that way in the template, but I find it more confusing to see alone like this, so might be better to show the full compiled line. We also need to show them the option of adding the WebAssembly interactive modes, which work as well.

```chsharp
builder.Services.AddRazorComponents()
    .AddInteractiveServerComponents();

// or

builder.Services.AddRazorComponents
    .AddInteractiveWebAssemblyComponents();

// or both

builder.Services.AddRazorComponents()
    .AddInteractiveServerComponents()
    .AddInteractiveWebAssemblyComponents();
```

and

```
app.MapRazorComponents<App>()
    .AddInteractiveServerRenderMode();

// or

app.MapRazorComponents<App>()
    .AddInteractiveWebAssemblyRenderMode()
    .AddAdditionalAssemblies(typeof(Counter).Assembly);

// or

app.MapRazorComponents<App>()
    .AddInteractiveServerRenderMode()
    .AddInteractiveWebAssemblyRenderMode()
    .AddAdditionalAssemblies(typeof(Counter).Assembly);
```

ReadMe.md
```diff
@@ -157,11 +160,18 @@ Or for the Pro version with additional features:
...
 8. Create a Razor Component page with a map:
 
    ```html
    @page "/"
+   @rendermode InteractiveServer
```
TimPurdum, line 174: No great way to show an inline comment here, so might put the other options into the text above or below, like:

8. Create a Razor Component page with a map. Be sure to set an [Interactive Render Mode](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes) like Server, WebAssembly, or Auto:

TimPurdum (approved):
Looks good, you can add those suggestions or re-assign back to me. The other copies of the getting started information are in the Pro repository, I can make sure those get updated.

TimPurdum:
@codergrl I went ahead and implemented my suggestions along with some other changes we have discussed. I also:

- updated the internal `ReadMe.md` that gets packaged for NuGet.org
- fixed the broken style at the top of the github readme
- updated the link to ArcGIS Locations Service
- figured out a better solution for the "old version docs" (the branch/tag dropdown in github).

codergrl:
This looks great, good to merge.

PR #408, opened 2025-05-21
Update actions

TimPurdum:
- `bumpVersion.ps1`: The `nuget search` command here apparently uses an old .NET Framework version of NuGet, which isn't supported on the GH runners. I replaced it with a web API query to find the latest published version of the package.
- `dev-pr-build.yml`: I'm not sure why I took the `npm build` step out of the dev build pipeline, but I think that was a bad idea, we want to know at PR time if we break something in the TypeScript layer. I've added it back in here. It's copied straight from the main release build, so it should be good to go.
- `main-release-build.yml`
   - Added comments for clarity
   - Flipped `GenerateDocs` to `true`. I'm pretty sure when we had set this to `false` in the past it would cause missing Intellisense in the NuGet packages.
   - I tested this on the `gh-action-test` branch`, and got it to pass, except for the parts that actually push to NuGet and create the Release. This helped me find the issues with the `bumpVersion` script.

### DIFF
--- a/.github/workflows/dev-pr-build.yml
+++ b/.github/workflows/dev-pr-build.yml
@@ -33,14 +33,25 @@ jobs:
       run: |
         ./bumpVersion.ps1
         
+    - name: Restore dependencies
+      run: dotnet restore ./src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
+        
     - name: Update NPM
       uses: actions/setup-node@v4
       with:
         node-version: '>=22.11.0'
         check-latest: 'true'
-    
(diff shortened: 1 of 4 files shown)


codergrl (approved):
This seems ok but I don't know how I would go about testing it.

TimPurdum:
image
@codergrl just so you know where this is, if you wanted to run a pipeline test, you can do so from the `Actions` tab. Of course, that's always a little scary if you don't know what's in the file! This one I set so that it will never publish unless it is actually running on the main branch, and then I ran it 3 times on the `gh-action-testing` branch.

PR #409, opened 2025-05-27
Pro/print widget

TimPurdum:
Mostly removes support types that are only used in GeoBlazor Pro


src/dymaptic.GeoBlazor.Core/assetCopy.ps1
```diff
@@ -1,7 +1,8 @@
-﻿$SourceFiles = "./node_modules/@arcgis/core/assets/"
+﻿$SourceFiles = Join-Path $PSScriptRoot "node_modules/@arcgis/core/assets/"
 
-$OutputDir = "./wwwroot/assets"
-$packageJson = (Get-Content "./package.json" -Raw) | ConvertFrom-Json
+$OutputDir = Join-Path $PSScriptRoot "wwwroot/assets"
```
AndersenBell, line 3: maybe add a comment that this will copy assets for both core or pro

TimPurdum: It's not actually copying them to both folders. It's still copying them in the Core folder, and Pro uses them from the core folder. This is just fixing the paths, because PowerShell relative paths are based on where you execute the script from, so this replaces the relative paths with absolute paths based on the script location.

PR #411, opened 2025-05-29
Feature/410 experimental layers

TimPurdum:
- Removed the `[Experimental]` attribute on `WFSLayer`
- Moved a bunch of classes from `Components` to `Models` as they appear to really be mostly read from the WFS server, not really settable from the client.
- In code-generated TypeScript, removed `WfsCapabilities` from copying methods for `WFSLayer` and `WFSLayerInfo`, as these capabilities are huge classes containing all the feature definitions, and they slow down serialization and are redundant.
- Other TS changes were mostly automatic by the code generator. `IQueryMixin` isn't really used, so I simplified it, but left it there so the code generator wouldn't re-generate it.
- Fixed a bug in `arcGisJsInterop.lookupGeoBlazorId` and `lookupGeoBLazorGraphicId` where `undefined` in `uid` would cause spamming the same GeoBlazor `Id` across many objects when serializing back to GB.
- Added some serialization tests
- Added a `WFSLayers.razor` samples pages, icon, and to nav menu.


seahro (approved):
Works well, one popup css issues, that is not related to the wfs page and functionality

PR #415, opened 2025-05-30
release cleanup

TimPurdum:
- Releasing the samples and templates was a challenge due to the combination of .net8 and .net9 in our repository, so this PR is moving all the samples and tests to .net9, leaving only the library itself targeting .net8 for backwards compatibility.
- Also removed the attempt to commit the published build version back to `main`. Since it's a protected branch, that fails. However, the way `bumpVersion.ps1` is written, this doesn't matter. The next time we release, it is going to query NuGet and bump to 4.0.1, unless we manually set it higher than that in the meantime.

### DIFF
--- a/.github/workflows/main-release-build.yml
+++ b/.github/workflows/main-release-build.yml
@@ -91,17 +91,6 @@ jobs:
             Exit 1
         }
         
-    - name: Add & Commit
-      # Only run this step if the branch is main
-      if: github.ref == 'refs/heads/main'
-      continue-on-error: true
-      run: |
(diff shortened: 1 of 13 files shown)


.github/workflows/main-release-build.yml
```diff
@@ -91,17 +91,6 @@ jobs:
             Exit 1
         }
         
-    - name: Add & Commit
```
seahro, line 94: see PR notes in conversation tab

seahro (approved):
looks good, also version bump in place from .5 to .6

PR #419, opened 2025-06-03
Bug/417 samples

### LINKED ISSUE #417: Samples Clean Up
- Fix console log error messages if possible
- Fix or remove broken source code links
- Fix flashing error on first load

### LINKED ISSUE #418: Error with ElementExpressionInfo in PopupContent
There is a Protobuf serialization error with `ElementExpressionInfo` in `PopupContent`. Apparently we did not create a serialization record for that type.

TimPurdum:
Closes #417 

- Removed broken links to source pages. I got this working for our samples site in the pro repository using our docs site as well, but it's not necessary for other users to see our samples here, and they can peruse source directly in this repo. This included removing files like `MauiFileProvider`, `SharedFileProvider`, `SharedDirectoryContents`, `SharedFileEnumerator`, `SourceCode.razor` etc., and calling code in program starting and `MainLayout.razor`.
- Fixed a bunch of the nasty navigation bugs that you get when navigating away while a map is rendering. Some of these fixes are in GeoBlazor proper, in `arcGisJsInterop`, by using catches better, but most of them are handled within the samples themselves, by using javascript in `functions.js` to register back to blazor when the map is rendered, so that blazor can block navigation during a map rendering, and then also call `destroy` on map components before navigating.

Also closes #418

- Added `ElementExpressionInfoSerializationRecord`, and the converting to/from methods with `ElementExpressionInfo`.
- Updated `PopupContentSerializationRecord` to use the correct type.
- Updated `graphic.json` definition, removing the `returnType`, since this is readonly and doesn't need to be serialized via protobuf.
- Added a unit test `TestPopupsWithFieldInfoContent` to validate changes. This test throws an exception before the changes.


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Shared/NavMenu.razor
```diff
@@ -62,13 +79,17 @@
 
         if (firstRender)
         {
+            IJSObjectReference? pro = await JsModuleManager.GetArcGisJsPro(JsRuntime, CancellationToken.None);
```
AndersenBell, line 82: could you add some comments as to why you have to initialize geoblazor here in the nav menu?

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Shared/NavMenu.razor
```diff
@@ -47,13 +57,20 @@
...
+    private async ValueTask LocationChanged(LocationChangingContext arg)
     {
+        int tries = 0;
+        while (!_viewReady && tries < 10)
+        {
+            tries++;
+            // Wait until the view is ready before disposing all views
+            await Task.Delay(100);
+        }
+        _ = await JsRuntime.InvokeAsync<bool>("disposeAllViews");
```
AndersenBell, line 72: add a comment here, why we need to call this, when the maps' disposal should be called when navigating away from the pages?

AndersenBell (approved):

PR #424, opened 2025-06-11
Add AuthenticationManager.RegisterToken, Update ReadMes

TimPurdum:
References Gissue 165 in the Pro repository

## AuthenticationManager

- Implemented the method `AuthenticationManager.RegisterToken` so developers can set up an [App Authentication](https://developers.arcgis.com/documentation/security-and-authentication/app-authentication/) flow. Implemented in C# with `DateTimeOffset` for the `expires` parameter to be strongly-typed, and converting to milliseconds when passing to TypeScript.
- In `authenticationManager.ts`, the `registerToken` method has to set the `server` property to a default for ArcGIS Online if the user hasn't set a `PortalUrl`. I did some testing (not directly in GeoBlazor, hoping to write a test shortly here), and if you don't set a server, it fails.
- Implemented `AuthenticationManager.TokenExpirationDateTime` and `GetTokenExpirationDateTime` so a dev could query the current token lifespan if necessary.

## ReadMe

- In both `ReadMe` files, I removed the "Getting Started" content and made a big link to the docs site getting started page instead. This should help us be able to keep that up to date and not let them slip out of sync.
- For the outer ReadMe that shows as the GitHub landing page, I replaced the top icon and removed some styling that wasn't working, so that they will show better when pushed to GitHub.


TimPurdum:
Addressing a few Claude comments here.

First, apparently Claude cannot see PR fixes, as it continues to flag issues even after you fix them, like missing typescript return types.

Second, as a pass-through library, a lot of the validation logic Claude is suggesting really belongs in either ArcGIS or the consuming application, not in GeoBlazor.

Third, while it is true that `ConfigureAwait(false)` is suggested for _general_ library code, this is not the case for code that is _specific_ to Asp.NET Core UI applications. Asp.NET Core does not normally have a Synchronization Context, and does not require this setting even in library code. See [here](https://devblogs.microsoft.com/dotnet/configureawait-faq/#i’ve-heard-configureawait(false)-is-no-longer-necessary-in-.net-core.-true). It is true that a consumer could in theory set up a synchronization context, then call the authentication manager methods, thereby introducing the overhead caused by not adding `ConfigureAwait(false)`, but this is an edge case IMO, and not worth dealing with at this time. Certainly it's not something we want to deal with in just one place, but would want to change across the library all at once.

Some of Claude's comments were excellent, I did end up adding the missing `getTokenExpiration` implementation, and I removed all the `try/catches` in the TS class, which I have done in other places as well, so that the developers using GeoBlazor can actually see the errors in .NET.

PR #426, opened 2025-06-12
Fix missing constructor tags for DynamicDataLayer, DynamicMapLayer for .NET 9

### LINKED ISSUE #425: Bug: Dynamic Data Layer / Map Layer failing to render
From https://samples.geoblazor.com/map-image-layers
- Click on `Show Toggleable Highways, Cities, Railroad Layers`
Image

Issue is missing `[ActivatorUtilitiesConstructor]`

TimPurdum:
Closes #425 

Since these two classes weren't code generated, they missed the .NET 9 added attribute `[ActivatorUtilitiesConstructor]`. This attribute is necessary starting in .NET 9 for Razor components that have two constructors.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <CoreVersion>4.0.1.0</CoreVersion>
+    <CoreVersion>4.0.1.1</CoreVersion>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


TimPurdum:
> Is there a build task you could add to check that this is set correctly on all the classes with multiple constructors?

Good idea. Not very straightforward, because it's not every class necessarily, just ones inheriting from `MapComponent`/`ComponentBase`. So probably would require quite a bit of Roslyn. I'll write a gissue for it.

PR #440, opened 2025-07-02
fix for legend in layerlist widget

### LINKED ISSUE #439: LayerListWidget Not able to Display Legend Content
Reported on Discord https://discord.com/channels/1027907220949717033/1389940047385727127
Looks like the TypeScript didn't get updated to match the changes in the C# schema with the new widget layers.

TimPurdum:
Closes #439 

# LayerListWidget

- Consolidated `ListItemPanelContent` into `ListItemPanelWidget`, which brings it more inline with what we had in V3, and less layers of abstraction to wade through to set up.
- `ListItemPanelWidget`
   - Moved constructors to hand-edited file
   - Removed `Content`
   - Added `StringContent`, `WidgetContent`, and `HtmlElementContent`
   - Added `ShowLegendContent` boolean. This was in V3, and matches the custom "legend" string logic in ArcGIS.
   - Updated the TypeScript implementation with actually building the widget content and to support the HTML element reference content. Notice, I didn't have to change the rest, as it was already expecting these content types to be on `ListItemPanelWidget`, not on the child `ListItemPanelContent`, which was the bug.
- Updated the `LayerLists.razor` sample to include legends in the widget.

# ESBuild

- Unrelated, I updated `esbuild.js` based on a discussion that @AndersenBell and I had, to trigger a new build when the git branch changes. We discovered that it was possible for the file date check to fail/think there were no changes on a branch change, presumably because git somehow caches and restores files.
- I also moved the recorded build output file outside of the project folder, since this file was _also_ triggering rebuilds in .net, defeating the whole purpose of the fewer builds in the first place.


AndersenBell:
The SQL query sample dropdown needs the style updated:
image

AndersenBell (approved):

TimPurdum:
@AndersenBell apologies, I did a 360 on this after you reviewed it.

I remembered the reason for the `ListItemPanelContent` class was to support _multiple_ content items in a single panel. However, it was definitely clunky.

The new implementation:
- `ListItemPanelWidget.Content` - just one property of type `IReadOnlyList<ListItemPanelContent>`, this is how V4 is now, so this is less of a breaking change.
- `ListItemPanelContent` - this is now a `record` with implicit conversions from string, Widget, and ElementReference, so a user can create a list of any combination of those without having to instantiate this class.
- Kept the new (restored from V3) `ListItemPanelWidget.ShowLegendContent`
- Updated the TypeScript implementation to match.
- Implemented unit tests for the 4 different content types.

PR #452, opened 2025-07-16
Updating WMSLayer test and typescript files

### LINKED ISSUE #446: Fix WMSLayerTests
The layer used in WMSLayerTests is not reliable, and fails to load often. Find a more reliable layer and/or figure out how to make the loading more reliable.

AndersenBell:
Closes #446 

Updates the WMS layer test to use a faster loading source and load a single layer

Updating the TypeScript files when calling `fetchImage` to return the JS reference.
Adding a call for `fetchImageAsString` to return a string that contains the image.


test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/TestRunnerBase.razor
```diff
@@ -143,10 +143,9 @@
         _testRenderFragments[methodName] = fragment;
     }
 
-    protected async Task WaitForMapToRender([CallerMemberName] string methodName = "")
+    protected async Task WaitForMapToRender([CallerMemberName] string methodName = "", int tries = 100)
```
TimPurdum, line 146: Nice! I was just thinking of the fact that we needed a way to adjust the timeout for different tests. I think the WebMap tests could use longer timeouts as well.

TimPurdum: Although, it would be nice to make it a clear timespan timeout instead of a `tries` property. Could we make the new parameter `timeoutInSeconds` or `timeoutInMilliseconds` and then do a conversion in this method to create the number of tries? Then the caller could just say `await WaitForMapToRender(timeoutInSeconds: 60)`.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/WMSLayerTests.razor
```diff
@@ -23,26 +23,30 @@
             Assert.AreEqual(layer?.Id, createdLayer.Id);
             layerViewCreated = true;
         }
-        AddMapRenderFragment(
-            @<SceneView @ref="sceneView"
-                                  class="map-view"
-                                  OnViewRendered="renderHandler"
-                                  OnLayerViewCreate="OnLayerViewCreated">
+
+        AddMapRenderFragment(@<SceneView @ref="sceneView"
```
TimPurdum, line 27: This is nitpicky, but I like the opening component on a new line so it lines up more like natural markup.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/WMSLayerTests.razor
```diff
@@ -23,26 +23,30 @@
...
                 </Map>
             </SceneView>);
         await WaitForMapToRender();
         int tries = 100;
+
         while (!layerViewCreated && tries > 0)
         {
             await Task.Delay(100);
             tries--;
         }
```
TimPurdum, line 45: We should probably also create a reusable `WaitForLayerViewCreated` method as well. I'm not sure how easy that would be, and doesn't need to be part of this PR, just thinking about it since you were refactoring the other wait method. But we definitely use this pattern a lot.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/WMSLayerTests.razor
```diff
@@ -23,26 +23,30 @@
...
-            @<SceneView @ref="sceneView"
-                                  class="map-view"
-                                  OnViewRendered="renderHandler"
-                                  OnLayerViewCreate="OnLayerViewCreated">
+
+        AddMapRenderFragment(@<SceneView @ref="sceneView"
+                                         class="map-view"
+                                         OnViewRendered="renderHandler"
+                                         OnLayerViewCreate="OnLayerViewCreated">
                 <Map>
```
TimPurdum, line 31: We should add `<SpatialReference Wkid="102100" />` to this SceneView, because it is giving the "squashed" globe currently, if you pause the test you can see it.

TimPurdum (changes requested):
Code looks great. If you can refactor that `WaitForMapToRender` parameter and add the SpatialRef, this will be ready to go.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/WMSLayerTests.razor
AndersenBell: Rider auto-formatting refuses to adhere to this

TimPurdum: Ah, that explains it at least, thanks.

Issue #441, opened 2025-07-02
Replace deprecated ArcGIS Widgets with WebComponents


TimPurdum:
ArcGIS Widgets are now mostly deprecated, and being replaced by web components. The plan is to maintain the external GeoBlazor API of our Widgets (which are closer to the ArcGIS web components anyways), and replace the internals with the web components. This task is to track that work. It might need to be split into multiple PRs or sub-gissues for different widgets.

PR #453, opened 2025-07-16
Move to Web Component-based MapView

TimPurdum:
This will be the _start_ of #441, but it's a big change and important enough to get right and do without also changing the widgets.

# Core C#
- `MapView.razor`
   - Added the `<arcgis-map>` and `<arcgis-scene>` web components directly in the Razor markup. They would have been a drop-in replacement for a `div`, except for the fact that I had to add the `if/else` logic for `SceneView`.
- `MapView.razor.cs`
   - `ShowZoomWidget` property - The web component maps do not have zoom controls by default, but since this would be a breaking change for GeoBlazor users, adding this property and setting it to `true` as default allows us to continue setting the zoom controls on all maps.
   - `_widgets`  collection - I updated several private field collections from `HashSet` to `List` because I realized that the order of widgets and layers was definitely important. In one of the samples, where we had multiple widgets along the top-right corner, the order was clearly off in the HashSet.
   - `_spatialReference` - removed, wasn't being used.
   - `DisposeAsync` - removed call to JS for disposal. Map Components come with `autoDestroy`. https://developers.arcgis.com/javascript/latest/references/map-components/arcgis-map/#autoDestroyDisabled
   - `OnAfterRenderAsync` - the AbortManager set here was duplicating the work in the base class `MapComponent`.
   - `RenderView` - added the `ZoomWidget` based on the new property, inserted as first in the widget list so it would always show in the top right.
   - `BuildMapView` - refactored this out to simplify the `SceneView` overloads.
- `SceneView.cs`
   - The `RenderView` method constantly was getting out of sync, so I created a simpler `BuildMapView` override that will be easier to keep in sync, and removed the `RenderView` override.
- `CustomOverlay.razor` - removed the `calcite-theme-light` class, which wasn't needed and conflicts when users select dark theme.
- `CustomOverlay.razor.css` - Added `z-index: 1`. This was a big surprise for me, it appears that by default with the web components my overlays no longer show on top, but just bumping up to 1 was enough to bring them back into view.
- `BingMapsLayer.cs` - Marked as obsolete with a clear warning that it is not expected to work.
- `Layer.cs` - Moved `FeatureLayer.ExcludeApiKey` to the base `Layer.cs` class, so it can be used on other layer types.

# Core TypeScript

- `arcGisJsInterop.ts`
   - Added imports for `map-components`, `calcite-components`, and `kernel.js` - to get the arcgis version.
   - Added imports for `ArcgisMap` and `ArcgisScene`, which are the JS/TS representations of the web components.
   - Removed some uncalled code
   - In `setAssetsPath`, I added logic that will prevent end users from getting errors when we upgrade ArcGIS assets. This is paired with changes in `assetCopy.ps1`. It also sets the path for assets for Calcite and MapComponents.
   - In `buildMapView` are the majority of refactorings.
      - We no longer need the "pre-load" of layers
      - Removed the call to dispose the old view (see link above about why)
      - We find the web component by id. If the users set no basemap and it's not a webmap/webscene, we have to add a basemap, otherwise it just won't load. This may be a _slightly_ visual breaking change if someone really had a sparsely set map in the past, as I believe with the older `View` you could get away with no basemap, even though it looked terrible. I chose the `light-gray` basemaps as the fallbacks, as they should be low-bandwidth and non-distracting.
      - I had to move setting the spatial reference earlier in the process for one of the samples to work. I believe that was the WMS sample where I kept getting the squashed globe without this change.
      - We keep the same logic to build Map, WebMap, or WebScene. Then we call `await mapComponent.viewOnReady()`, which waits for the map to generate the view. Then we just set our `view` variable to `mapComponent.view`, and the rest of the code basically works!
   - There's a bit more cleanup in this file, but it should all make sense. `disposeView` was able to be completely removed.
- `layer.ts` - moved the `excludeApiKey` logic here from `featureLayer.ts`.
- `widget.ts` - added the missing `zoom` case in the jsBuilder function.

# Build Process Files

- `esbuild.js` - Added a `--force` flag option, which can be called via `npm run forceDebugBuild`
- `package.json` 
   - Added that `forceDebugBuild` script definition
   - Upgraded the arcgis package ref
   - Added the map-components and calcite-components package refs
- `tsconfig.json` - Not really sure _exactly_ what these changes functionally change, but they follow ESRI suggestions and make sense (we are using a bundler, and moving to ES2023 seems reasonable).
- `assetCopy.ps1` - This one has a lot of changes
   - In addition to ArcGIS core package assets, it also copies assets for the map-components and calcite-components packages
   - It adds a version-named folder in the `wwwroot/assets` output path to avoid caching issues when we deploy upgrades.
   - It copies the `themes` files back to the _unversioned_ folder path, since all our users have this in a `<link>` tag and it would break if we moved them. These files are all CSS and I don't think are causing the breaking issues anyways.
   - I think I was able to solve the riddle why we had the "Move assets to root" extra step. Apparently, there is a bug in Powershell that alters the behavior based on if the target folder exists or not https://github.com/PowerShell/PowerShell/issues/21105. By pre-creating each target folder, I can hopefully guarantee consistency and remove that step. This was never an issue for me but for other devs, so we'll know if/when it breaks for someone else.

# Tests

- All the changes here are not really related to the branch, but are removing asserts from callback methods, where they were throwing into GeoBlazor catches that would not propagate them to the test runner, and throwing not for a good reason (they threw because there were more than one layer type hitting the methods). Replaced with a simpler type check.

# Samples

- Removed the `BingMapsLayers.razor` sample
- Added `ExcludeApiKey` to the `WCSLayers.razor` example. This probably wasn't actually necessary, because this is the one with the slow/broken server response that gives partial rendering, but it didn't hurt either.
- I realized that in one of my previous refactoring rounds, we ended up loading the arcgis theme _after_ `site.css`. This made it hard to style widgets and panels. So you will see several changes around that ordering.
   - `MainLayout.razor` - add the `site.css` here for MAUI apps so it can load after everything else. I don't _think_ the MAUI apps support `HeadContent`?
   - `Core.Sample.Maui/wwwroot/index.html` - removed `site.css`
   - `Core.Sample.Wasm/wwwroot/index.html` - moved `site.css` down into the body. The `<HeadContent />` tag _does_ appear to work here, but it gets injected at the very end of the head tag. Also in this file, fixed the favicon and loading icon.
- `NavMenu` - split out code-behind, tried to add more scroll fixes for that first-load double-bounce that can happen if you try to scroll right after load. Haven't solved it yet, but the structure is there to handle it.
- `functions.cs` - related scroll functions, also removed `checkMapViewReady`, as this is no longer needed!
- `site.css` - mostly UI css fixes, primarily for the `UpdateFeatureAttributes` sample in Pro.

As always, there will be a related Pro PR as well!


TimPurdum:
Closing, will re-open with unbundled assets code

PR #455, opened 2025-07-22
Bug/client testing bugs

TimPurdum:
This addresses several issues that arose while working with a customer.

# MapColor

- There was a bug, when creating a `MapColor` from a 9-digit Hex code (i.e. `#RRGGBBAA`). The System.Drawing.Color conversion utility we were using did not support this format, only 4 and 7 digit hex codes.
- Added a pre-check for the 9-digit version, and added custom parsing logic.
- Added a unit test class with conversion tests for MapColor.

# ExtentGeometryTests

- This is related to a bug, but I have not resolved or been able to reproduce it yet. In the client application, `GeometryEngine.GetExtentCenter(extent)` is throwing an exception when the extent calls back to `MapComponent.OnJSComponentCreated`. However, that does not reproduce in this test, so I still have to investigate, but I figured this was a gap regardless and good to add.

# Sketch/VisibleElements Scripts

- These changes are related to the Pro PR on the same branch name. Several TS classes were redefined to be inline with their C# classes, and moved into Pro.
 
Also, the `IAttributeBinsQuery` was updated by the CodeGeneration tool.


TimPurdum:
> Missing Input Validation in MapColor.cs:192-202:
> Issue: No validation that HexOrNameValue is actually 9 characters before calling Substring(7, 2)
> Risk: Potential ArgumentOutOfRangeException if hex string is malformed
> Fix: Add length check before substring operation

I'm not sure what Claude means here. The quoted code `if` statement is clearly checking `Length == 9`.

PR #462, opened 2025-07-31
fixing LocationService bug, adding support for single address string

### LINKED ISSUE #461: LocationService address object properties are incorrectly named
A user reported an issue with the `LocationService.AddressesToLocation` function only returning the general city location rather than an exact location of an address.

The ESRI documentation is incorrect in specifying the parameters of the javascript function call, the Address object needs an "Address" property rather than a "Street" property to correctly provide the location lookup.

Image

AndersenBell:
Closes #461 

Updates the Address class used by the `LocationService` class send the correct parameters to the service.
Adds the overloads to `LocationService` to support searching by a full address string

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Components/LocationService.cs
+++ b/src/dymaptic.GeoBlazor.Core/Components/LocationService.cs
@@ -270,12 +270,7 @@ public class LocationService : LogicComponent
     {
         IJSStreamReference streamRef = await InvokeAsync<IJSStreamReference>("addressesToLocations", url, addresses,
             countryCode, categories, locationType, outSpatialReference, requestOptions);
-        await using Stream stream = await streamRef.OpenReadStreamAsync(1_000_000_000L);
-        using var ms = new MemoryStream();
-        await stream.CopyToAsync(ms);
-        ms.Seek(0, SeekOrigin.Begin);
-        string json = Encoding.UTF8.GetString(ms.ToArray());
-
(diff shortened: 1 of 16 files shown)


src/dymaptic.GeoBlazor.Core/Components/LocationService.cs
```diff
@@ -152,133 +157,1101 @@ public async Task<List<AddressCandidate>> AddressesToLocations(List<Address> add
...
-        IJSStreamReference streamRef = await InvokeAsync<IJSStreamReference>("addressesToLocations", url, addresses, countryCode, categories, locationType, outSpatialReference, requestOptions);
-        await using Stream stream = await streamRef.OpenReadStreamAsync(1_000_000_000L);
-        using var ms = new MemoryStream();
-        await stream.CopyToAsync(ms);
-        ms.Seek(0, SeekOrigin.Begin);
-        string json = Encoding.UTF8.GetString(ms.ToArray());
+        IJSStreamReference streamRef = await InvokeAsync<IJSStreamReference>("addressToLocations", url, address,
+            categories, countryCode, forStorage, location, locationType, magicKey, maxLocations, outFields,
+            outSpatialReference, searchExtent, requestOptions);
+        var json = await ConverJsStreamToString(streamRef);
```
TimPurdum, line 1238: Same issue here, we broke this one by inserting an expected parameter in the middle of the JS parameters.

PR #470, opened 2025-08-21
Bug/456 authentication manager

magmoe:
Fixed issue with Authentication Manager redirecting to login based on incorrect url and set up the Token Refresh sample to test this. 

Closes #456

PR #473, opened 2025-09-11
Refactor Authentication Token Refresh Examples

TimPurdum:
I realized after approving and merging #470 that there were still some things that didn't make sense, so this is a follow-up refactoring of that work.

Changes:

- `AuthenticationApis.cs`
   - Greatly simplified the API implementations to make it a cleaner example.
   - I couldn't figure out why this file refused to let us use `Results.Ok` without the inline namespace. I tinkered around a bunch and never figured it out.
- `ArcGisAuthServiceWasm` - This class was confusing because it was being passed to both the wasm _and_ the _server_ page! It also had a weird combination of API communication and display string creating logic. I deleted the class and moved the logic into the pages and the new `WasmAuthService`
- `IAuthService` - Extracts a shared interface from `ArcGisAuthService` for client/server injection of the services
- `WasmAuthService` 
   - Client implementation of `IAuthService`
   - Simply passes the call to HttpClient and returns the response
- `MapPage(s).razor`
   - Added a third page for `InteractiveAuto` render mode
   - Moved all the string generation code into the pages. It's a bit redundant between the 3 pages, but I think it's cleaner than mixing this with the auth services.
- `Client.Program.cs` - removed the passing of `ArcGISApiKey` - we shouldn't ever need this.


seahro (approved):
tested refactors, without the ArcGISApiKey, and only the Geoblazor License key.  
- tested against a portal item on ArcGIS Online using an ArcGIS AppId and ArcGIS Client Secret that were registered on ArcGIS Online.
image

- tested against a portal item on our ArcGIS Enterprise using an AppId and Client Secret registered on our ArcGIS Enterprise/Portal.
image
image

both server and webassembly versions both connect successfully to their respective targets using the respective credentials in web assembly, server, and auto modes

PR #474, opened 2025-09-15
Added featurelayer tests

### LINKED ISSUE #472: Add full test coverage to FeatureLayer and FeatureLayerView methods
1. Collect a list of all the methods (including property async getter/setter methods) and `EventCallbacks` in the following files:
    - `FeatureLayer.cs`
    - `FeatureLayer.gb.cs`
    - `FeatureLayerView.cs`
    - `FeatureLayerView.gb.cs`
2. Compare the list of methods and events against the tests in `FeatureLayerTests.cs` and `FeatureLayerViewTests.cs`.
3. Write new tests to cover all the methods and events that are not currently covered. If it makes sense, you may combine multiple methods into a single test, for example a setter and a getter of a property, but ensure that the JavaScript code is being exercised (sometimes GeoBlazor will return the cached C# value even if the JS call fails) and checked, by using the `AssertJavaScript` pattern and the `testRunner.js` file. You can also create a `featureLayerAsserts.js` or `featureLayerViewAsserts.js` file to help expand these asserts, see the examples from the legend and layerList widgets.
4. It is quite possible that you will uncover bugs during this testing. Some usual issues/solutions:
    -  The C# method calls a `get{PropertyName}` method in JavaScript, but that method is not defined in the TypeScript wrapper class. Check the following files/types:
       - `featureLayer.ts` - `FeatureLayerWrapper`
       - `featureLayer.gb.ts` - `FeatureLayerGenerated`
       - `featureLayerView.ts` - `FeatureLayerViewWrapper`
       - `featureLayerView.gb.ts` - `FeatureLayerViewGenerated`
   - The TypeScript method called exists, but it is not property converting a type value between ArcGIS and C#. Typically this shows up as returning a raw value from the ArcGIS types (anything that is a complex object) or passing a complex object method parameter directly into an ArcGIS method without conversion. The solution is usually to look for a `buildJs{ObjectType` or `buildDotNetObjectType` method. If that does not exist, use `sanitize(dotNetObject)` to pass to ArcGIS from C# and `removeCircularReferences(jsObject)` to pass back to C#. Builders should be imported inline using the `let {buildJsObjectType} = await import('./objectType');` pattern, while the other two functions can be imported at the top of the file from `arcGisJsInterop`.
   - If a test fails in Server mode but passes in Wasm, and it has a large data load, then it likely needs to use a stream response rather than an object response to avoid the SignalR data size limit.
     - in the TS method, after getting the data, call `buildEncodedJson(data)` and return the result.
     - the C# method should call JS something like this:
        ```csharp
        IJSStreamReference jsStreamRef = await JsComponentReference!.InvokeAsync<IJSStreamReference>("matchingJsMethod");
        ExpectedDataType? result = await jsStreamRef.ReadJsStreamReference<ExpectedDataType>();
        ```

Don't forget, you can rebuild just the JavaScript portion of GeoBlazor by running `./esBuild.ps1` from the Core project folder and refreshing the browser. This does not require stopping the test runner or recompiling the C#.

magmoe:
I expanded our end-to-end test coverage around FeatureLayer and stood up the JavaScript assertion helpers. The FeatureLayer test suite now exercises real rendering of a map/layer, visibility toggling, adding/updating/deleting features via `ApplyEdits` and `Add`, and the core query flows (`QueryFeatures`, `QueryObjectIds`, `QueryRelatedFeatures`, and `QueryTopFeatures`). It also validates key metadata accessors such as capabilities and field/domain inspection, and includes a scenario that confirms we can use a public FeatureLayer without an API key. On the view side, I established baseline coverage for `FeatureLayerView` by adding a query test. To support all of this, I added small JS helpers (`featureLayerAsserts.js` and `featureLayerViewAsserts.js`) that the Razor tests call through `AssertJavaScript`, so each test asserts the state of the underlying ArcGIS JS objects. Net result: broader, more reliable coverage of our FeatureLayer interop paths with minimal harness changes.

Closes #472

### DIFF
--- a/test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
+++ b/test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
@@ -470,4 +470,359 @@
         await AssertJavaScript("assertLayerExists", args: "tile");
         await AssertJavaScript("assertLayerExists", args: "feature");
     }
+
+    [TestMethod]
+    public async Task FeatureLayer_BasicPropertyRoundtrip_And_JS_Asserts(Action renderHandler)
+    {
+        FeatureLayer? layer = null;
+
+        AddMapRenderFragment(
+            @<MapView class="map-view" OnViewRendered="renderHandler">
+                <Map>
+                    <Basemap>
+                        <BasemapStyle Name="BasemapStyleName.ArcgisTopographicBase" />
+                    </Basemap>
+                    <FeatureLayer @ref="layer"
+                                  Source="@([])"
+                                  OutFields="@(["*"])"
+                                  SpatialReference="SpatialReference.Wgs84"
+                                  GeometryType="FeatureGeometryType.Point"
+                                  ObjectIdField="OBJECT_ID" />
+                </Map>
+            </MapView>);
+
+        await WaitForMapToRender();
+        await AssertJavaScript("assertLayerExists", args: "feature");
+
+        await layer!.SetLegendEnabled(false);
+        Assert.AreEqual(false, await layer.GetLegendEnabled());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "legendEnabled", false]);
+
+        await layer!.SetPopupEnabled(true);
+        Assert.AreEqual(true, await layer.GetPopupEnabled());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "popupEnabled", true]);
+
+        await layer!.SetLabelsVisible(false);
+        Assert.AreEqual(false, await layer.GetLabelsVisible());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "labelsVisible", false]);
+
+        await layer!.SetReturnM(true);
+        Assert.AreEqual(true, await layer.GetReturnM());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "returnM", true]);
+
+        await layer!.SetReturnZ(true);
+        Assert.AreEqual(true, await layer.GetReturnZ());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "returnZ", true]);
+
+        await layer!.SetDisplayFilterEnabled(false);
+        Assert.AreEqual(false, await layer.GetDisplayFilterEnabled());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "displayFilterEnabled", false]);
+
+        await layer!.SetEditingEnabled(false);
+        Assert.AreEqual(false, await layer.GetEditingEnabled());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "editingEnabled", false]);
+
+        await layer!.SetUseViewTime(false);
+        Assert.AreEqual(false, await layer.GetUseViewTime());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "useViewTime", false]);
+
+        await layer!.SetScreenSizePerspectiveEnabled(false);
+        Assert.AreEqual(false, await layer.GetScreenSizePerspectiveEnabled());
+        await AssertJavaScript("assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "screenSizePerspectiveEnabled", false]);
+
+        await layer!.SetMinScale(5000);
+        Assert.AreEqual(5000, await layer.GetMinScale());
+        await AssertJavaScript("assertFeatureLayerNumberPropEquals", args: [layer!.Id, "minScale", 5000]);
+
+        await layer!.SetMaxScale(0);
+        Assert.AreEqual(0, await layer.GetMaxScale());
+        await AssertJavaScript("assertFeatureLayerNumberPropEquals", args: [layer!.Id, "maxScale", 0]);
+
+        await layer!.SetRefreshInterval(0.01);
+        Assert.AreEqual(0.01, await layer.GetRefreshInterval());
+        await AssertJavaScript("assertFeatureLayerNumberPropEquals", args: [layer!.Id, "refreshInterval", 0.01]);
+
+        await layer!.SetDefinitionExpression("1=1");
+        Assert.AreEqual("1=1", await layer.GetDefinitionExpression());
+        await AssertJavaScript("assertFeatureLayerStringPropEquals", args: [layer!.Id, "definitionExpression", "1=1"]);
+
+        await layer!.SetDateFieldsTimeZone("UTC");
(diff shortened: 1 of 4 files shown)


test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerViewTests.razor
```diff
@@ -58,4 +58,194 @@
...
+                                  SpatialReference="SpatialReference.Wgs84"
+                                  GeometryType="FeatureGeometryType.Point"
+                                  ObjectIdField="OBJECTID" />
+                </Map>
+            </MapView>);
+
+        await WaitForMapToRender();
+
+        // Wait for the FeatureLayerView
+        for (var i = 0; layerView is null && i < 100; i++) { await Task.Delay(50); }
```
TimPurdum, line 105: We have `WaitForLayerToRender` built in to the test runner.

```csharp
LayerViewCreateEvent createEvent = await WaitForLayerToRender<FeatureLayer>();
```

Whenever possible, try to avoid using the random timeouts. I know it's not always possible, but since we have this one it is.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerViewTests.razor
```diff
@@ -58,4 +58,194 @@
...
+
+        // ---------------------------
+        // Roundtrips for properties that CAN be set from C#
+        // ---------------------------
+
+        // Visible (bool) — set, then Get, and assert on live JS object
+        await layerView!.SetVisible(false);
+        var visible = await layerView.GetVisible();
+        Assert.AreEqual(false, visible);
+        await AssertJavaScript("assertObjectHasPropertyWithValue", args: [layerView.Id, "visible", false]);
```
TimPurdum, line 116: good!

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerViewTests.razor
```diff
@@ -58,4 +58,194 @@
...
+        await layerView.SetVisible(true);
+        visible = await layerView.GetVisible();
+        Assert.AreEqual(true, visible);
+        await AssertJavaScript("assertObjectHasPropertyWithValue", args: [layerView.Id, "visible", true]);
+
+        // HighlightOptions (object) — set & ensure roundtrip (object equality via C#)
+        var hl = new HighlightOptions(color: new MapColor("cyan"), fillOpacity: 1.0, haloOpacity: 1.0);
+        await layerView.SetHighlightOptions(hl);
+        var gotHl = await layerView.GetHighlightOptions();
+        Assert.IsNotNull(gotHl);
```
TimPurdum, line 127: Is there a reason we can't assert this one in JS?

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerViewTests.razor
```diff
@@ -58,4 +58,194 @@
...
+
+        // ---------------------------
+        // Highlight handles
+        // ---------------------------
+
+        var handle1 = await layerView.Highlight(graphics[0]);
+        Assert.IsNotNull(handle1);
+
+        var handle2 = await layerView.Highlight(new ObjectId(1));
+        Assert.IsNotNull(handle2);
```
TimPurdum, line 167: Can we also test the `handle.Remove()` method?

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerViewTests.razor
```diff
@@ -58,4 +58,194 @@
...
+        var hl = new HighlightOptions(color: new MapColor("cyan"), fillOpacity: 1.0, haloOpacity: 1.0);
+        await layerView.SetHighlightOptions(hl);
+        var gotHl = await layerView.GetHighlightOptions();
+        Assert.IsNotNull(gotHl);
+
+        // Optional: feature effect as an object (do not set IncludedEffect here)
+        var fx = new FeatureEffect();
+        await layerView.SetFeatureEffect(fx);
+        var gotFx = await layerView.GetFeatureEffect();
+        Assert.IsNotNull(gotFx);
```
TimPurdum, line 249: The only part of this test that isn't a duplicate of the previous test is the last four lines. Can you reorganize to remove the duplication?

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
```diff
@@ -470,4 +470,361 @@
...
+    await eventLayer!.Add(new Graphic(new Point(-117, 34, 0), new SimpleMarkerSymbol(color: new MapColor("black"))));
+
+    await eventLayer.Refresh();
+    var refreshEvent = await Task.WhenAny(refreshTcs.Task, Task.Delay(5000));
+    Assert.AreSame(refreshTcs.Task, refreshEvent);
+    }
+
+
+    [TestMethod]
+    public async Task FeatureLayer_All_Getters_Invoke_JS(Action renderHandler)
```
TimPurdum, line 740: Let's rename this something like `HostedFeatureLayer_Can_Call_Getters_On_All_Properties`

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
```diff
@@ -470,4 +470,361 @@
...
+                                  ObjectIdField="OBJECTID" />
+                </Map>
+            </MapView>);
+
+        await WaitForMapToRender();
+        await AssertJavaScript("assertLayerExists", args: "feature");
+
+        await layer!.SetLegendEnabled(false);
+        Assert.AreEqual(false, await layer.GetLegendEnabled());
+        await AssertJavaScript("featureLayerAsserts.assertFeatureLayerBooleanPropEquals", args: [layer!.Id, "legendEnabled", false]);
```
TimPurdum, line 499: Looking back at this file now, we could simplify and move all of these to `assertObjectHasPropertyWithValue`.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/FeatureLayerTests.razor
```diff
@@ -470,4 +470,361 @@
...
+                </Map>
+            </MapView>);
+
+        await WaitForMapToRender();
+        await AssertJavaScript("assertLayerExists", args: "feature");
+
+        await layer!.SetOutFields(new[] { "Field1", "OBJECTID", "Field2" });
+
+        var outFields = await layer.GetOutFields();
+        CollectionAssert.AreEquivalent(new[] { "Field1", "OBJECTID", "Field2" }, outFields?.ToArray());
```
TimPurdum, line 591: There might be a few like these collection asserts that won't work with `assertObjectHasPropertyWithValue` and need to keep the custom functions in `featureLayerAsserts.js`.

TimPurdum (changes requested):
Getting close! Mostly cleaning up now.

PR #476, opened 2025-09-16
`activatorutilitesconstructor` attribute validation console app

seahro:
adds the console app to the build process that checks to ensure that all classes with 2+constructors have the required `ActivatorUtilitesConstructor` attribute.
-Claude Generated the main console app.
-tested the build process before and after finding and resolving the criteria that is targeted by the app...
Example Output before resolving the missing attributes:

Example Output after adding in the missing attributes
image

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
+++ b/src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
@@ -114,4 +114,12 @@
         <Message Importance="high" Text="Clearing locks on esBuild files..." />
         <Exec Command="pwsh $(ProjectDir)../../esBuildClearLocks.ps1" ContinueOnError="false" />
     </Target>
+
+    <Target Name="ValidateConstructors" AfterTargets="Build">
+        <Message Importance="high" Text="Validating constructor patterns..." />
+        <Exec Command="dotnet run --project $(ProjectDir)../../test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/dymaptic.GeoBlazor.Core.Test.ConstructorTesting.csproj --configuration $(Configuration)"
+              WorkingDirectory="$(ProjectDir)"
+              ContinueOnError="true"
+              IgnoreExitCode="true" />
+    </Target>
 </Project>
--- a/src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
+++ b/src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
@@ -117,9 +117,10 @@
 
     <Target Name="ValidateConstructors" AfterTargets="Build">
         <Message Importance="high" Text="Validating constructor patterns..." />
-        <Exec Command="dotnet run --project $(ProjectDir)../../test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/dymaptic.GeoBlazor.Core.Test.ConstructorTesting.csproj --configuration $(Configuration)"
+        <Exec Command="$(ProjectDir)../../test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/bin/$(Configuration)/net9.0/dymaptic.GeoBlazor.Core.Test.ConstructorTesting.exe"
               WorkingDirectory="$(ProjectDir)"
               ContinueOnError="true"
-              IgnoreExitCode="true" />
+              IgnoreExitCode="true"
+              Condition="Exists('$(ProjectDir)../../test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/bin/$(Configuration)/net9.0/dymaptic.GeoBlazor.Core.Test.ConstructorTesting.exe')" />
     </Target>
 </Project>
--- a/test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
+++ b/test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
@@ -1,0 +1,145 @@
+using Microsoft.Build.Locator;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.MSBuild;
+
+namespace dymaptic.GeoBlazor.Core.Test.ConstructorTesting;
+
+class Program
+{
+    static async Task<int> Main(string[] args)
+    {
+        // Register MSBuild
+        MSBuildLocator.RegisterDefaults();
+
+        // Find the solution file
+        var currentDirectory = Directory.GetCurrentDirectory();
+        var solutionPath = FindSolutionFile(currentDirectory);
+
+        if (string.IsNullOrEmpty(solutionPath))
+        {
+            Console.WriteLine("Could not find solution file (.sln) in the repository.");
+            return 1;
+        }
+
+        using var workspace = MSBuildWorkspace.Create();
+        var solution = await workspace.OpenSolutionAsync(solutionPath);
+
+        var nonCompliantClasses = new List<(string className, string filePath)>();
+
+        foreach (var project in solution.Projects)
+        {
+            var compilation = await project.GetCompilationAsync();
+            if (compilation == null) continue;
+
+            foreach (var syntaxTree in compilation.SyntaxTrees)
+            {
+                var semanticModel = compilation.GetSemanticModel(syntaxTree);
+                var root = await syntaxTree.GetRootAsync();
+
+                // Find all class declarations
+                var classDeclarations = root.DescendantNodes().OfType<ClassDeclarationSyntax>();
+
+                foreach (var classDeclaration in classDeclarations)
+                {
+                    var classSymbol = semanticModel.GetDeclaredSymbol(classDeclaration);
+                    if (classSymbol == null) continue;
+
+                    // Check if the class inherits from ComponentBase
+                    if (!InheritsFromComponentBase(classSymbol)) continue;
+
+                    // Get all constructors
+                    var constructors = classSymbol.Constructors
+                        .Where(c => !c.IsImplicitlyDeclared && c.DeclaredAccessibility == Accessibility.Public)
+                        .ToList();
+
+                    // Check if class has 2 or more constructors
+                    if (constructors.Count >= 2)
+                    {
+                        // Check if there's a parameterless constructor with [ActivatorUtilitiesConstructor]
+                        var hasCompliantParameterlessConstructor = constructors.Any(c =>
+                            c.Parameters.Length == 0 &&
+                            HasActivatorUtilitiesConstructorAttribute(c));
+
+                        if (!hasCompliantParameterlessConstructor)
+                        {
+                            var filePath = syntaxTree.FilePath;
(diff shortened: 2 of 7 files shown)


src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
```diff
@@ -114,4 +114,13 @@
         <Message Importance="high" Text="Clearing locks on esBuild files..." />
         <Exec Command="pwsh $(ProjectDir)../../esBuildClearLocks.ps1" ContinueOnError="false" />
     </Target>
+
+    <Target Name="ValidateConstructors" AfterTargets="Build">
+        <Message Importance="high" Text="Validating constructor patterns..." />
+        <Exec Command="$(ProjectDir)../../test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/bin/$(Configuration)/net9.0/dymaptic.GeoBlazor.Core.Test.ConstructorTesting.exe"
+              WorkingDirectory="$(ProjectDir)"
+              ContinueOnError="true"
+              IgnoreExitCode="true"
```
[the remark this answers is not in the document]
TimPurdum: Actually, I agree with Copilot on this, @seahro. You'll notice in `Directory.Build.props` I have `TreatWarningsAsErrors` set to `true`. The reasoning is that ignored warnings are a major cause of technical debt and require later refactoring. In this case, it's all the more important because these components will fail at runtime for users if they are missing the expected constructor and are used as Razor Components. Please change to `ContinueOnError="false"`.

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
```diff
@@ -114,4 +114,13 @@
         <Message Importance="high" Text="Clearing locks on esBuild files..." />
         <Exec Command="pwsh $(ProjectDir)../../esBuildClearLocks.ps1" ContinueOnError="false" />
     </Target>
+
+    <Target Name="ValidateConstructors" AfterTargets="Build">
+        <Message Importance="high" Text="Validating constructor patterns..." />
+        <Exec Command="$(ProjectDir)../../test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/bin/$(Configuration)/net9.0/dymaptic.GeoBlazor.Core.Test.ConstructorTesting.exe"
```
TimPurdum, line 120: Several items:

1. This path assumes that the test project has been compiled, and if not, it skips this completely. If you instead called `dotnet run` on the test project, then it would not need to be pre-compiled.
2. Since this is really part of our core build process, I don't think it belongs in the `test` folder or should be called a test. I would like the new project to be named `repoRoot/src/dymaptic.GeoBlazor.Core.Validator`. This generic name could then be expanded with any other logic necessary, and also be called from the Pro repository as well.
3. It doesn't _really_ matter as long as we break the build (see my other comment on the Copilot comment), but I would place this as `BeforeTargets="Build"`.
4. Pass the Core project path to the new project as a command line argument. See my notes in your `Program.cs` for how to use that.

suggested refactor:

```xml
<Target Name="ValidateConstructors" BeforeTargets="Build">
    <Message Importance="high" Text="Validating constructor patterns..." />
    <Exec Command="dotnet run dymaptic.GeoBlazor.Core.Validator.csproj $(MSBuildProjectFullPath)"
            WorkingDirectory="$(ProjectDir)../dymaptic.GeoBlazor.Core.Validator"
            ContinueOnError="false"
            IgnoreExitCode="false" />
</Target>
```

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+
+        return projectViolations;
+    }
+
+    /// <summary>
+    /// Analyzes a single project for constructor compliance.
+    /// </summary>
+    static async Task<List<(string className, string filePath)>> AnalyzeProject(Project project, string solutionPath)
+    {
+        var violations = new List<(string className, string filePath)>();
```
TimPurdum, line 75: Replace this tuple with a record:

```csharp
public record ViolatingClass(string ClassName, string FilePath);
```

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+        var fileCount = syntaxTrees.Count;
+        var currentFile = 0;
+
+        foreach (var syntaxTree in syntaxTrees)
+        {
+            currentFile++;
+            var fileName = Path.GetFileName(syntaxTree.FilePath);
+
+            // Show progress - using carriage return to overwrite the same line
+            Console.Write($"\rAnalyzing file {currentFile}/{fileCount}: {fileName,-50}");
```
TimPurdum, line 89: `\r` is not a standard line break character. Windows is `\r\n` and Unix is `\n`. Typically for cross-platform support we use `$"{Environment.NewLine}"`

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+        return projectViolations;
+    }
+
+    /// <summary>
+    /// Analyzes a single project for constructor compliance.
+    /// </summary>
+    static async Task<List<(string className, string filePath)>> AnalyzeProject(Project project, string solutionPath)
+    {
+        var violations = new List<(string className, string filePath)>();
+        var compilation = await project.GetCompilationAsync();
```
TimPurdum, line 76: By getting the full project compilation, does this return each class re-combined, `.cs` and `.gb.cs`? Have you double-checked that this is reading data from both locations?

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+        // Skip classes that don't inherit from ComponentBase
+        if (!InheritsFromComponentBase(classSymbol)) return null;
+
+        // Get public constructors (excluding compiler-generated ones)
+        var constructors = classSymbol.Constructors
+            .Where(c => !c.IsImplicitlyDeclared && c.DeclaredAccessibility == Accessibility.Public)
+            .ToList();
+
+        // Only check classes with 2 or more constructors
+        if (constructors.Count < 2) return null;
```
TimPurdum, line 150: This should be `< 1`. It would also be a mistake to _only_ have a parameterized constructor (unless that's already a compilation error? I can't recall).

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+            // Check for MapComponent first (most common in this codebase)
+            if (typeName == "MapComponent")
+            {
+                return true;
+            }
+
+            // Check for ComponentBase (base class of MapComponent and other Blazor components)
+            if (typeName == "ComponentBase")
+            {
+                return true;
```
TimPurdum, line 236: I think I would like this to throw an exception. I can't think of a scenario where we want non-MapComponent Razor Components in GeoBlazor.

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /// <summary>
+    /// Checks if a constructor has the [ActivatorUtilitiesConstructor] attribute.
+    /// </summary>
+    static bool HasActivatorUtilitiesConstructorAttribute(IMethodSymbol constructor)
+    {
+        return constructor.GetAttributes().Any(attr =>
+            attr.AttributeClass?.Name == "ActivatorUtilitiesConstructorAttribute" ||
+            attr.AttributeClass?.Name == "ActivatorUtilitiesConstructor");
```
TimPurdum, line 252: Surely one of these two lines is wrong and can be removed? I know it's confusing to know when it is with or without the `Attribute` postfix, but I would think it should be consistent on testing.

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
TimPurdum: Actually, the better check here would be

```csharp
if (constructors.All(c => c.Parameters.Length == 0))
{
    // no parameterized constructors, so we don't need the attribute on a parameterless constructor
    return null;
}
```

since the whole reason for the attribute is to differentiate the parameterless constructor from one _with_ parameters.

TimPurdum (changes requested):
Good start!

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
seahro: makes sense, refactored

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
seahro: missed that ... removed the incorrect attribute value

test/dymaptic.GeoBlazor.Core.Test.ConstructorTesting/Program.cs
seahro: -missed that greater than 1, refactored. 
-compiler will catch all instances where parameterless constructors are missing
-agree that we just need to target the parameterless contructor without attributes

src/dymaptic.GeoBlazor.Core/dymaptic.GeoBlazor.Core.csproj
seahro: -initially i decided on the bypassing any build warnings, because i didnt want the validator to stop other development.  but understand the reason for keeping enforcement universal.  Refactored.
also adjusted the name and readme to explain how to use this to validate either repo.  will be updating the pro repo pr with this also vs duplicating the console app.

seahro:
Claude's feedback:
first 2 items were cleaned up.  Good catch.
the 3rd item:  The logic does 

  Line 117-119: We get ALL public constructors (both parameterless and parameterized)

  Line 121: if (constructors.All(c => c.Parameters.Length == 0))
  - This checks if we have ONLY parameterless constructors
  - If true (meaning we have 1+ parameterless but NO parameterized), we return false - no
  violation
  - If false (meaning we have at least one parameterized constructor), we continue

  Line 126: We filter to get just the parameterless constructors

  Line 134: return parameterlessConstructors.Count > 0
  - Returns true only if there's a parameterless constructor
  - But we already know from line 121 that there's ALSO at least one parameterized constructor

  So yes, the logic ensures we have more than 1 constructor total, specifically:
  - At least 1 parameterless constructor
  - At least 1 parameterized constructor

  This means we're checking for classes that have multiple constructors of different types (both
   parameterless and parameterized), and flagging them if the parameterless one lacks the
  attribute.

  Classes with:
  - Only 1 parameterless constructor → NOT flagged (line 121 returns false)
  - Only parameterized constructors → NOT flagged (line 134 returns false)
  - Both types with attribute → NOT flagged (line 128 returns false)
  - Both types without attribute → FLAGGED (line 134 returns true)

PR #482, opened 2025-10-02
remove portal url manipulation and update documentation

### LINKED ISSUE #480: Enterprise Portal Url Cannot be hard-coded to include `/portal`
We discovered while working with a user on Discord that it is possible to change the default Enterprise url paths so that they might not end with `/portal`. So we need to remove the logic we just added that will add this, and instead update the documentation to explain that the full url is up to the user to get correct in the app settings.

TimPurdum:
Closes #480 

There will also be a matching PR in Pro for documentation, please review both. 

# Changes

- Removed `AuthenticationManager.NormalizePortalUrl`. We are going to make it the responsibility of the user to set the correct portal url, since not all Enterprise portals end with `/portal`.
- Updated the documentation in `ReadMe.md` and `Home.razor` in the `TokenRefresh` sample project to explain that Enterprise portals _typically_ end with `/portal`, but not always. Also updated the AGOL portion to point out that the token refresh query works against both `arcgis.com` and `your_company.arcgis.com`.

# Testing

- Run the `TokenRefresh` WebApp sample with dymaptic Enterprise app credentials. Use the url `https://arcgis.dymaptic.com/portal` for `ArcGISPortalUrl`. Verify that the token queries and maps work in Wasm and Server modes.
- Run the `TokenRefresh` WebApp sample with dymaptic AGOL app credentials. Use the url `https://moraveclabsllc.arcgis.com` for `ArcGISPortalUrl`. Verify that the token queries and maps work in Wasm and Server modes.
- Run the `TokenRefresh` WebApp sample with dymaptic AGOL app credentials. Use the url `https://arcgis.com` for `ArcGISPortalUrl`. Verify that the token queries and maps work in Wasm and Server modes.
- Run the Test runner `dymaptic.GeoBlazor.Core.Test.WebApp` and the `AuthenticationManagerTests` using the same urls and tokens as above.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <CoreVersion>4.2.1.0-beta-2.1</CoreVersion>
+    <CoreVersion>4.2.1.0-beta-3</CoreVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <CompressionEnabled Condition="$(Configuration) != 'RELEASE' OR $(EnableCompression) != 'true'">false</CompressionEnabled>
     <DefineConstants Condition="$(EnableCompression) == 'true'">$(DefineConstants);ENABLE_COMPRESSION</DefineConstants>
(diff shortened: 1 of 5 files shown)


seahro (approved):
pages all updated to use the url as entered. successful in token refresh for portal and arcgis. then after getting an initial error, i corrected the ago url to  moraveclabs.maps.arcgis.com and worked.
image
image
image

Tests Pass
image

PR #485, opened 2025-10-14
Add loadProtobuf to Layer.Load

### LINKED ISSUE #484: Error on queries of Loaded layers when no map is used
Users are supposed to be able to call `layer.Load` to initialize a layer and then perform queries on the data without rendering it in a map.

I believe this used to work, however, at some point, probably when we added protobuf serialization, this broke. When querying for features, we now get this error:

```
Cannot read properties of undefined (reading 'fromObject') TypeError: Cannot read properties of undefined (reading 'fromObject') at getProtobufGraphicStream (https://localhost:7249/_content/dymaptic.GeoBlazor.Core/js/core_chunk_REA7WVAV.js:39465:45) at FeatureLayerWrapper.queryFeatures (https://localhost:7249/_content/dymaptic.GeoBlazor.Core/js/core_featureLayer_KRF6T33J.js:702:24)
at dymaptic.GeoBlazor.Core.Components.Layers.FeatureLayer.QueryFeatures(Query query, CancellationToken cancellationToken)
```

It turns out the `undefined` is the `ProtoGraphicsCollection`, because we are missing the call to `loadProtobuf` in `arcGisJsInterop`.

Temporary workaround for users:

```csharp
var coreModule = await JsModuleManager.GetArcGisJsCore(JsRuntime, null, CancellationToken.None);
await coreModule.InvokeVoidAsync("loadProtobuf");
```

Resolution for issue: add a call to `await coreModule.InvokeVoidAsync("loadProtobuf");` to `Layer.Load()` in `Layer.cs`.

TimPurdum:
Closes #484 

Adds a call to `arcGisJsInterop.loadProtobuf` from `Layer.Load` in `Layer.cs`, which allows queries of features to succeed in serializing and returning the graphics.

Also added a test, which will fail if you comment out the new line in `Layer.Load`.

**UPDATE**

This PR was expanded a bit after more feedback from the user who reported the initial bug.

New changes:

- In `Layer.Load`, we now retrieve back the loaded layer and copy the properties into the C# instance, so that they are populated. This required moving `load()` from the generated files to the root files in TypeScript for _all_ layer types, so you will see a lot of TS file changes, but they are all relatively equivalent. There were slight cleanup items done when in those files based on linting highlighting.
- Updated `Layer.UpdateFromJavaScript` and removed a lot of the copying logic. This method has always been mostly redundant since we added `MapComponent.CopyProperties`. We still need it for the Sublayers, but the user reported that `Fields` were getting duplicated, and you can see why in the code I deleted in `FeatureLayer.UpdateFromJavaScript`.

### DIFF
--- a/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
+++ b/src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
@@ -2068,6 +2068,10 @@ export let ProtoGraphicCollection;
 export let ProtoViewHitCollection;
 
 export async function loadProtobuf() {
+    if (ProtoGraphicCollection && ProtoViewHitCollection) {
+        // already loaded
+        return;
+    }
     load("_content/dymaptic.GeoBlazor.Core/graphic.json", function (err, root) {
         if (err) {
             throw err;


seahro (approved):
wanted to investigate the issues that claude had found.  and its directly related to what you mentioned about removing the unnecessary `UpdateFromJavaScript` method.  and switching over to the `MapComponent.CopyProperties`
While running the tests with and without layer.load I do like the suggestion that is given.
image

Issue #487, opened 2025-10-27
Add @bind-Value to all Parameters


TimPurdum:
Using the Code Generator, add the `{ParameterName}Changed` Parameter to all current non-`EventCallback` parameters. This exposes 2-way binding to end users, who can then use `@bind-ParameterName="_localVariable"` to bind. 

Example:

```csharp
public class DistanceMeasurement2DWidget: Widget
{

...

    // existing parameter
    [Parameter]
    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
    public DistanceMeasurement2DViewModel? ViewModel { get; set; }

    // new parameter
    [Parameter]
    [JsonIgnore]
    public EventCallback<DistanceMeasurement2DViewModel> ViewModelChanged { get; set; }
}
```

Usage:

```html
<MapView>
  <Map>
  </Map>
  <DistanceMeasurement2DWidget @ref="_distanceWidget" 
                                     Visible="false" 
                                     Position="OverlayPosition.BottomRight"
                                     @bind-ViewModel="_distanceWidgetViewModel"
                                     @bind-ViewModel:after="StateHasChanged" />
</MapView>
```

PR #488, opened 2025-10-27
measurement widgets and sample

### LINKED ISSUE #486: DistanceMeasurement2DWidget & AreaMeasurement2DWidget not rendering
Several issues:
- EnumToKebabCase converter - the `2D` digit is messing up the converter and adding a null char
- ViewModels are not being loaded, but they are required for some functionality of these widgets.

TimPurdum:
Closes #486 

# Problem 1

The serialization of the two widgets, `DistanceMeasurement2DWidget` and `AreaMeasurement2DWidget` was failing.

- Fixed the hyphen placement on `area-measurement-2d` in `widget.ts`
- Updated `StringExtensions.ToKebabCase`.
  - Currently on `develop`/`main`, if you add a `DistanceMeasurement2DWidget`, and put a breakpoint in the TypeScript code, you will see that the serialized value has a unicode `null` value as the last character. This is because the cound of `upper` is thrown off by the special logic for numbers. 
  - I also realized that, while the `string.Create` method is in theory fastest, the fact that I was pre-scanning the string with `val.Count(char.IsUpper)` somewhat defeats the purpose. So I replaced it with a simpler for loop over the char array with a StringBuilder. This removes the need to track offsets or worry about the length.
 
# Problem 2

When I went to add the `MeasurementWidgets.razor` sample, and based on the discussion in Discord, I realized that there actually _is_ a way that we can expose 2-way binding to all parameters in GeoBlazor. I am spinning that off into a separate task with #487. However, I went ahead and implemented it as a proof-of-concept for these two widgets and their view models here.

- Implemented `ViewModelChanged` in both Widget classes.
- Updated `MapComponent.CopyProperties` and `MapComponent.CopyProperty` to find and trigger any bound `Changed` parameters after the ArcGIS values are returned.
- Created the sample, svg icon, and added to nav menu.
- Fixed the `RenderModeSelector` in the Core Samples WebApp to show `Auto` mode by default, rather than `Static`.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <CoreVersion>4.3.0.6</CoreVersion>
+    <CoreVersion>4.3.0.7</CoreVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <CompressionEnabled Condition="$(Configuration) != 'RELEASE' OR $(EnableCompression) != 'true'">false</CompressionEnabled>
     <DefineConstants Condition="$(EnableCompression) == 'true'">$(DefineConstants);ENABLE_COMPRESSION</DefineConstants>


TimPurdum:
Changes made while fixing:

- Simplified the sample
- Fixed/updated `Widget.SetContainerId` and `Widget.SetPosition`
- Expanded the `OverlayPosition` enum with the new values that are available in ArcGIS
- Added tests for `Widget.SetContainerId`/`Widget.SetPosition` and the new viewmodel binding
- Consolidated a testRunner.js function

AndersenBell (approved):
re-tested and approved

PR #489, opened 2025-10-27
Feature/open basemaps

TimPurdum:
1. This PR makes it optional to add an ArcGIS API Key/AppID to get started with GeoBlazor.
    - `MapView.razor.cs`
       - Remove the ApiKey checking logic and current default basemap logic in `RenderView`
       - Added a new `EnsureBasemap` method that will add an `OpenStreetMapLayer` if no other layer is provided.
       - Updated `BuildMapView` to pass some new parameters to JavaScript.
       - Updated some references.
   - `ITileLayer.cs` - Added this new custom interface to make it simpler to identify tile layers that can serve as basemaps.
   - `OpenStreetMapLayer`/`WebTileLayer` - FullExtent should have been readonly on these layers.
2. While doing that, it made sense to also take the opportunity to rename and separate some tightly-coupled ArcGIS and GeoBlazor logic.
    - Created `geoBlazorCore.ts`
       - split out much of the non-ArcGIS-specific code from `arcGisJsInterop` into this new file.
       - added an `export * from './arcGisJsInterop';`, which acts as a pass-through, so anything that used to reference `arcGisJsInterop` can now reference `geoBlazorCore` and still work.
       - This caused renaming of imports in all the other TS files, so there are a _lot_ of TypeScript file changes, but most of them can be ignored.
    - I also renamed `JsModuleManager.GetArcGisCore` and `JsModuleManager.GetArcGisPro`
3. If we are supporting users starting without the API key, then I wanted to control the experience when they attempt to add an ArcGIS element. So the logic in `geoBlazorCore.buildMapView` with `observers` is there to listen for the ESRI auto-login popup and override/hide it. Instead, we show a message in the map view. You can test the difference by setting `MapView.AllowDefaultEsriLogin`, which will show the ESRI modal.
4. The test runner broke significantly with trying to run tests without the API key, followed by tests _with_ the API key. So I had to rebuild a _lot_ of it. 
    - The basic change is to create iFramed `[IsolatedTest]` classes that don't share a global authentication configuration with the rest of the tests. But this required rewriting a ton of code throughout the `Core.Test.Blazor.Shared` library. I'm happy for feedback on the changes, but I think the most important piece is whether or not it works for our purposes, since it's not really user-facing like the library.
    - `TestWrapper.razor` - new wrapper component, abstracts both isolated and inline test runner classes
       - In this wrapper, I'm using a library, `BlazorFrame`, which wraps `iframe`, and handles sending and receiving messages.
    - `TestFrame.razor` - this is the new "page" that the iframe navigates to and shows the isolated test class.
       - `testFrame.razor.js` - the `BlazorFrame` library only handled sending/receiving as the parent window, so I copied a bunch of their implementation to create the frame code itself and communicate back to the window.
   - `Index.razor` - I updated this class to track and store progress in `sessionStorage`, so that if you refresh the browser on a crash, you don't lose your test results. I also am passing the cancellation tokens into the test classes now, to stop things more quickly.
   - I updated the nuget packages, because I was getting an error, and apparently the `ThrowsExceptionAsync` was replaced with `ThrowsAsync`.
   - I disabled the WebAssembly debugger in `Core.Test.WebApp` `launchSettings.json`, because it was breaking on trying to understand the iframes.

Unrelated Dev-Experience Changes:
- Added some fixes for the optional `esBuildLogger` to `ESBuildLauncher.cs` and `esBuildLogger.ps1`
- Updated some conditionals in `Core.csproj` to try to make hot reload more responsive (still a work in progress)
- Created `.editorconfig` files, copying my settings from the Rider/Resharper `DotSettings` files, so that Visual Studio would have more similar rules to Rider.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <CoreVersion>4.3.0.7</CoreVersion>
+    <CoreVersion>4.3.0.8</CoreVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <CompressionEnabled Condition="$(Configuration) != 'RELEASE' OR $(EnableCompression) != 'true'">false</CompressionEnabled>
     <DefineConstants Condition="$(EnableCompression) == 'true'">$(DefineConstants);ENABLE_COMPRESSION</DefineConstants>


samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Basemaps.razor
```diff
@@ -1,20 +1,33 @@
 ﻿@page "/basemaps"
 
 <PageTitle>Basemaps</PageTitle>
-<h1>BaseMaps</h1>
+<h1>Basemaps</h1>
 <div class="links-div">
     <a class="btn btn-secondary" target="_blank" href="https://developers.arcgis.com/javascript/latest/display-a-custom-basemap-style/">ArcGIS API for JavaScript Reference</a>
 </div>
-<p class="instructions">Use the radio toggle to explore the 3 different ways to load a Basemap.</p>
+<p class="instructions">
```
AndersenBell: When I was testing this, I removed my ArcGIS key from both dymaptic.GeoBlazor.Core.Sample.WebApp and dymaptic.GeoBlazor.Core.Sample.WebApp.Client.
I saw the prompt to enter an ArcGIS key on the top of running page.
 
I was able to load the ESRI basemaps even without entering a key and changing the rendering mode.

After I entered a bogus key, it worked as expected.

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/NoArcGisApiKeyTests.razor
```diff
@@ -0,0 +1,27 @@
+﻿@inherits TestRunnerBase
```
AndersenBell: This aslo threw an error

```
Assert.ThrowsExactlyAsync failed. Expected exception type:. Actual exception type:. 'action' expression: '() => WaitForMapToRender(timeoutInSeconds: 4)'.
at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.<>c__DisplayClass225_1`1[[System.TimeoutException, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].b__1(String userMessage) in /_/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs:line 541
at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.d__223`1[[System.TimeoutException, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].MoveNext() in /_/src/TestFramework/TestFramework/Assertions/Assert.ThrowsException.cs:line 486
at dymaptic.GeoBlazor.Core.Test.Blazor.Shared.Components.NoArcGisApiKeyTests.TestArcGISBasemapWithNoApiKeyShowsError(Action renderHandler) in C:\repo\GeoBlazor\test\dymaptic.GeoBlazor.Core.Test.Blazor.Shared\Components\NoArcGisApiKeyTests.razor:line 25
at dymaptic.GeoBlazor.Core.Test.Blazor.Shared.Components.TestRunnerBase.RunTest(MethodInfo methodInfo) in C:\repo\GeoBlazor\test\dymaptic.GeoBlazor.Core.Test.Blazor.Shared\Components\TestRunnerBase.razor:line 401
```

test/dymaptic.GeoBlazor.Core.Test.Blazor.Shared/Components/AuthenticationManagerTests.cs
```diff
@@ -73,65 +73,42 @@ public class AuthenticationManagerTests: TestRunnerBase
     public async Task TestRegisterOAuthWithArcGISPortal()
```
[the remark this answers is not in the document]
TimPurdum: There are instructions to make those tests pass in that file at the top.

samples/dymaptic.GeoBlazor.Core.Sample.Shared/Pages/Basemaps.razor
TimPurdum: This all sounds like bad cached data in the browser. I can't reproduce this either. Although it was a good thing to test, as I only get the UI element I want with BasemapStyles, not with the portal Ids and tile layers (since those don't trigger the login popup).


https://github.com/user-attachments/assets/0decd3dc-c514-48d9-b362-89e6177f983c

TimPurdum:
New changes after PR review:

- Added/Updated Exception handling in `MapView.razor.cs` and `AuthenticationManager.cs` so that end users can listen for asynchronous errors if they wish by providing a `Func<Exception Task<bool>>` to handle the error.
- Added a `logUncaughtError` function in `geoBlazorCore.ts`, and hooked it up in `arcGisJsInterop.buildArcGisMapView` and `authenticationmanager.constructor` to the `esriConfig.logs.interceptors. This catches the asynchronous failures that do not get caught by our inline try/catches.
- Removed unnecessary async/awaits in TypeScript logic to make error handling smoother.
- Fixed the "reset/retry" logic in `geoBlazorCore.ts` and `arcGisJsInterop.ts` so that when one map load fails, the next mapview is reset correctly. Main logic is in `arcGisJsInterop.resetMapComponent`.
- Fixed the UI of the map error messaging to be generic and move the css to `geoblazor.css` to make sure it is applied regardless of how/where it is added.

@AndersenBell @seahro this is ready for re-review and testing. Specifically, I tested the `Basemaps.razor` sample without any ArcGIS key, and all the basemaps work except the `BasemapStyle` ones, which now show a clearer message and don't break further basemap loading.

AndersenBell (approved):

PR #521, opened 2026-06-08
V4.5.0

TimPurdum:
This pull request introduces several infrastructure and documentation improvements focused on standardizing the development/build environment, improving cross-platform support, and clarifying project structure and workflows. The most significant changes include major updates to the build system for better ESBuild integration and locking, expanded and reorganized documentation in `CLAUDE.md`, and new configuration files for Docker and Git. There are also updates to code style and editor configuration.

**Build System and Cross-Platform Improvements:**

* Major overhaul of `Directory.Build.props` to support platform-specific build tools, atomic ESBuild locking, and improved property visibility for MSBuild; adds logic to select the correct toolchain per OS/architecture and ensures ESBuild runs only once per build.
* Updates `Directory.Build.targets` to add a new `RemoveStaleJsContent` target, preventing build errors from stale JS/map files after ESBuild runs, and disables file watching on generated/script/static assets to improve performance.

**Documentation and Project Structure:**

* Expands and reorganizes `CLAUDE.md` with a detailed project structure diagram, clearer build/test instructions, TypeScript build workflow, version management, and information about RAG semantic search and specialized Claude agent configurations. [[1]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7L1-R19) [[2]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7R35-R122) [[3]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7L118-R171)

**Configuration and Environment Standardization:**

* Adds a `.dockerignore` file to exclude development, build, and secret files from Docker contexts, improving container build hygiene.
* Adds a `.gitattributes` file to enforce consistent line endings (LF for scripts/configs), reducing cross-platform issues.

**Code Style and EditorConfig Updates:**

* Updates `.editorconfig` (renamed and moved to repo root) to enforce explicit type declarations (`var` usage disabled), prefer file-scoped namespaces, and removes a redundant naming rule for private constants; also adds XML formatting rules for MSBuild files. [[1]](diffhunk://#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR7-R11) [[2]](diffhunk://#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bL97-R104) [[3]](diffhunk://#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bL133-R138) [[4]](diffhunk://#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bL243-L246)


TimPurdum (commented):
## 🤖 Code Review by Claude (Opus 4.8)

_This review was generated by Claude Code at Tim's request and posted via his account. It is advisory — not a blocking gate._

### Scope
This is a large release PR (**644 files, +24.6k / −142k**). The vast majority of the churn is generated output (`*.gb.cs` / `*.gb.ts`), file deletions/renames, docs, and config. I focused a line-level review on the **hand-written, correctness- and security-sensitive** code:

- TypeScript interop: `arcGisJsInterop.ts`, `geoBlazorCore.ts`, `baseComponent.ts` (the new `sanitize()` path + theme handling)
- Source generators: `SerializationGenerator`, `ProcessHelper`, `ESBuildGenerator`, `ProtobufDefinitionsGenerator`
- Serialization: `UnionConverter`, `MapColor`
- Build/lock infrastructure: `ProcessRunner`, `ProcessKiller`, `AcquireBuildLock`
- Analyzer: `ParameterConversionAnalyzer`

Generated files, bulk deletions, docs, and `.editorconfig`/`.gitattributes`/`.dockerignore` were **not** line-reviewed.

### Overall
The infrastructure work (atomic build locking, cross-platform tool selection, the protobuf serialization generator, the new `sanitize()` recursion) is solid and well-documented. I found **one likely runtime regression** (theme switching to light) and several **robustness issues** in the build/generator utilities that are worth addressing before or shortly after merge. Details are inline; lower-priority notes below.

### Inline findings (by severity)
1. **High** — `setComponentTheme` adds `calcite-mode-dark` regardless of theme and never removes it (light-theme regression). `arcGisJsInterop.ts`
2. **Medium** — `ProcessHelper.Execute` retry has no max-attempt bound (potential infinite retry); the process-kill match never hits. 
3. **Medium** — `ESBuildGenerator.ClearESBuildLocks` fire-and-forget `Task.Run` uses `SourceProductionContext` after the generator callback returns.
4. **Medium** — `ProcessKiller` substring match + kill-all can terminate unrelated processes (e.g. `dotnet`).
5. **Medium** — `ProcessRunner.RunCommand` leaves the child process running when the `CancellationToken` fires.
6. **Low/Medium** — `MapColor.GetHashCode` is inconsistent with `Equals` (array reference hash) — now matters more as a `record`.
7. **Low** — `SerializationGenerator` mis-parses a non-generic `Task` return type.

### Lower-priority notes (no clean anchor)
- **`SnappingOptions.gb.cs`**: this PR removes the public `Set*` / `AddTo*` / `RemoveFrom*` setter methods from the generated output. If that's an intentional generator change, great — but it is a **public API reduction** for anyone who called those setters directly. Worth a release-note line.
- **`UnionConverter.cs`**: `using System.Text.Unicode;` appears unused. In the collection `Number` branch, `reader.GetInt32()` is called before confirming an int property exists and will throw on a non-int/overflowing number. In `sanitize()` (`geoBlazorCore.ts`), the `layerId`→`layer` rewrite adds a `layer` key while iterating the same object with `for...in`; relying on V8 not revisiting added keys is fragile (and you would not want to recursively `sanitize()` a live ArcGIS layer).
- **`ParameterConversionAnalyzer`**: `HasImplicitConversionFromString` only inspects members declared on the target type, so an inherited implicit `string` operator would be missed. The diagnostic message's escaped fix-it string is hard to read.
- **`SerializationGenerator`**: `SerializableMethodRecord.ReturnType` property is `string?` but the positional ctor param is non-nullable `string` and is always passed a non-null value, so the `ReturnType != null` void-method branch is effectively dead code.

Thanks for the thorough infra work here — the locking and generator pieces are a real improvement.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -179,19 +181,37 @@ export function setTheme(theme: string | null, viewId): string | null {
...
+            if ((component instanceof MapView || component instanceof SceneView)) {
+                component!.container!.style.colorScheme = theme as string;
+                if (!component!.ui.container!.classList.contains('calcite-mode-dark')) {
+                    // if the view was already rendered, this class is missed and needs adding
+                    component!.ui.container!.classList.add('calcite-mode-dark');
+                }
+            }
+
+            if (component instanceof HTMLElement) {
+                component.classList.add('calcite-mode-dark');
```
TimPurdum: **🤖 Claude review — High: light-theme regression.**

`setComponentTheme` now adds `calcite-mode-dark` **unconditionally**, for both the `MapView`/`SceneView` branch (line ~205) and this `HTMLElement` branch (line 210). The previous `setViewTheme` only added the class `if (theme === 'dark')`.

Two consequences:
- When `setTheme('light', …)` is called, the light CSS link is loaded and `colorScheme` is set to `light`, but `calcite-mode-dark` is still added → Calcite widgets render dark on a light map.
- There is no code path that *removes* `calcite-mode-dark`, so toggling dark→light at runtime leaves widgets stuck in dark mode.

Suggest gating on the theme and removing the class otherwise, e.g.:
```ts
const isDark = theme === 'dark';
component.ui.container!.classList.toggle('calcite-mode-dark', isDark);
// and for the HTMLElement branch:
component.classList.toggle('calcite-mode-dark', isDark);
```

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/ProcessHelper.cs
```diff
@@ -0,0 +1,447 @@
...
+                    }
+                    catch
+                    {
+                        // ignore
+                    }
+                }
+
+                await Task.Delay(500);
+
+                await Execute(processName, workingDirectory, fileName, shellArguments, context, showConsole, sessionId,
```
TimPurdum, line 245: **🤖 Claude review — Medium: unbounded retry + dead kill path.**

This recursive `Execute(...)` re-runs whenever stdout/stderr contains `"The process cannot access the file"`. There is **no max-attempt counter**, so if the file lock persists the method retries forever (every ~500 ms), each time logging an error against `context`.

Also, the kill that is meant to clear the lock can't match: `Process.GetProcessesByName("dotnet")` + `p.MainModule?.FileName.Contains(programName)` compares the script name (`shellArguments[1]`, e.g. `ESBuild.cs`) against the path to `dotnet.exe`, which never contains the script name — so `runningProc` is essentially always `null` and the kill is a no-op.

Suggest threading a `int attempt = 0` (or a max-retries parameter) through `Execute` and bailing after N attempts, and revisiting how the offending process is actually located (command-line args, not `MainModule.FileName`).

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/ESBuildGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    private static void ClearESBuildLocks(ESBuildGeneratorOptions options,
+        SourceProductionContext context)
+    {
+        // only remove stale files during a full build
+        string[] args = ["ESBuildClearLocks.dll", "--stale-files"];
+
+        _ = Task.Run(async () => await ProcessHelper.Execute("Clear Locks",
```
TimPurdum: **🤖 Claude review — Medium: fire-and-forget escapes the generator callback.**

This `_ = Task.Run(async () => await ProcessHelper.Execute(..., context, ...))` is not awaited, but the captured `SourceProductionContext` (used inside `Execute` for `ReportDiagnostic` and `CancellationToken`) is only valid **for the duration of this `RegisterSourceOutput` callback**. By the time the spawned task runs `Execute`, the callback has likely already returned, so diagnostics reported from the task may throw or be silently dropped, and `context.CancellationToken` is no longer meaningful.

Relatedly, the `"Cleared ESBuild Process Locks"` log on line 92 fires immediately — before the lock-clearing process has actually run — so the message is misleading.

If the intent is to not block the generator, consider doing the work synchronously (generators already accept some latency) or moving lock-clearing into an MSBuild target rather than the generator.

build-tools/utilities/ProcessKiller.cs
```diff
@@ -0,0 +1,41 @@
...
+    public static bool KillByProcessOrFileName(string fileName)
+    {
+        Process[] processes = Process.GetProcesses();
+        bool killed = false;
+
+        foreach (Process process in processes)
+        {
+            try
+            {
+                if (process.ProcessName.Contains(fileName, StringComparison.OrdinalIgnoreCase))
```
TimPurdum, line 17: **🤖 Claude review — Medium: substring match can kill unrelated processes.**

`process.ProcessName.Contains(fileName, OrdinalIgnoreCase)` (and the `Modules` filename `Contains` below) is a substring match against **every** process on the machine, and every match is killed. Called with a common token like `"dotnet"` or `"node"`, this would also terminate the developer's IDE build server, other running .NET apps, etc.

Consider an exact (case-insensitive) name match, and/or restricting to processes the build itself launched (tracked PIDs). Enumerating `process.Modules` for all processes is also expensive and throws `Win32Exception` for protected/elevated processes (currently swallowed).

build-tools/utilities/ProcessRunner.cs
```diff
@@ -0,0 +1,270 @@
...
+                                    failureTriggerWord = triggerWord;
+                                }
+                            }
+                        }
+                    };
+
+                    process.BeginOutputReadLine();
+                    process.BeginErrorReadLine();
+
+                    await process.WaitForExitAsync(cancellationToken);
```
TimPurdum, line 155: **🤖 Claude review — Medium: child process orphaned on cancellation.**

`await process.WaitForExitAsync(cancellationToken)` throws `OperationCanceledException` when the token fires, which unwinds past the `process.Kill(true)` on line 164 — so a cancelled build leaves the spawned `dotnet`/`pwsh` process (and its children) running. Consider a `try/finally` that kills the process tree if `!process.HasExited` on the way out.

Minor, same method: `failureTriggered` / `lineWasEmpty` are mutated from the `OutputDataReceived` and `ErrorDataReceived` handlers, which fire on separate threadpool threads — unsynchronized shared state. And the XML doc on both `RunDotnetCommand` overloads says "without capturing output", but output is captured.

src/dymaptic.GeoBlazor.Core/Model/MapColor.cs
```diff
@@ -84,99 +99,152 @@ public double[]? RgbaValues
...
+
+            return new MapColor((IReadOnlyList<double>)blendedValues);
+        }
+
+        return null;
     }
 
     /// <inheritdoc />
+    [ArcGISMethod]
     public override int GetHashCode()
```
TimPurdum, line 192: **🤖 Claude review — Low/Medium: GetHashCode inconsistent with Equals.**

`HashCode.Combine(RgbaValues, HexOrNameValue)` hashes `RgbaValues` by **array reference**, but `Equals` compares it by value (`SequenceEqual`). Two `MapColor`s that are `Equals`-equal (same components, different array instances) can therefore produce different hash codes, which breaks them as `Dictionary`/`HashSet` keys.

This was latent before, but it matters more now that `MapColor` is a `record` (value-equality semantics are implied and records are commonly used as keys). Suggest hashing the contents, e.g. fold the `RgbaValues` elements into the `HashCode`, or hash `ToHex()`.

(Separately: the new `implicit operator string?` combined with `implicit operator MapColor(string)` creates bidirectional implicit conversions, which can surprise overload resolution and string interpolation — worth confirming that's intended.)

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/SerializationGenerator.cs
```diff
@@ -0,0 +1,607 @@
...
+        List<SerializableMethodRecord> methodRecords = [];
+
+        foreach (MethodDeclarationSyntax method in methods)
+        {
+            string returnType = method.ReturnType.ToString();
+
+            if (returnType.StartsWith("Task") || returnType.StartsWith("ValueTask"))
+            {
+                int bracketIndex = returnType.IndexOf('<');
+                returnType = returnType.Substring(bracketIndex + 1, returnType.Length - bracketIndex - 2);
```
TimPurdum, line 544: **🤖 Claude review — Low: non-generic `Task` return type mis-parsed.**

When `returnType` is a bare `Task` or `ValueTask` (no `<…>`), `IndexOf('<')` returns `-1`, so `Substring(bracketIndex + 1, returnType.Length - bracketIndex - 2)` becomes `Substring(0, Length - 1)`, producing garbage (e.g. `"Task"` → `"Tas"`). Guard for `bracketIndex < 0` and treat it as void/no-return. (Today this is masked because `ReturnType` is effectively always non-null and serialized methods return `Task<T>`, but it's a sharp edge for any future `[SerializedMethod]` that returns a non-generic `Task`.)

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/ProcessHelper.cs
TimPurdum: @copilot fix this issue.

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/ESBuildGenerator.cs
TimPurdum: @Copilot fix this issue

src/dymaptic.GeoBlazor.Core/Model/MapColor.cs
TimPurdum: @Copilot fix this issue

src/dymaptic.GeoBlazor.Core/Scripts/featureSnappingLayerSource.gb.ts
```diff
@@ -12,6 +12,19 @@ export async function buildJsFeatureSnappingLayerSourceGenerated(dotNetObject: a
     if (hasValue(dotNetObject.enabled)) {
         properties.enabled = dotNetObject.enabled;
     }
+
+    if (hasValue(dotNetObject.layerId)) {
+        if (arcGisObjectRefs.hasOwnProperty(dotNetObject.layerId)) {
+            properties.layer = arcGisObjectRefs[dotNetObject.layerId];
+        } else {
+            let dnLayer = await dotNetObject.dotNetComponentReference.InvokeMethodAsync("GetLayerForJs");
```
TimPurdum: Copilot/GPT-5.5: `DotNetObjectReference` exposes `invokeMethodAsync`, not `InvokeMethodAsync`. This fallback path runs when the referenced layer is not already in `arcGisObjectRefs`, so feature snapping with a layer that must be fetched from .NET will throw at runtime instead of building the source. Please switch this to the lower-case JS interop API, ideally null-safe: `dotNetObject.dotNetComponentReference?.invokeMethodAsync("GetLayerForJs")`.

build-tools/build-scripts/ESBuild.cs
```diff
@@ -423,188 +456,260 @@ static void CopyScriptsToPro(string coreScriptsDir, string proScriptsDir, bool v
...
+/// (from the most recent build) are preserved. This avoids the dotnet/sdk#49988 race condition.
 /// </summary>
 /// <param name="recordFilePath">Path where the record should be saved.</param>
 /// <param name="branch">The current Git branch name.</param>
-static void SaveBuildRecord(string recordFilePath, string branch)
+/// <param name="previousTimestamp">The timestamp from the prior build record (shifted to previousTimestamp).</param>
+static void SaveBuildRecord(string recordFilePath, string branch, long previousTimestamp)
 {
     // buffer 30 seconds into the future to avoid edge cases
     long timestamp = DateTimeOffset.UtcNow.AddSeconds(30).ToUnixTimeMilliseconds();
```
TimPurdum: Copilot/GPT-5.5: Persisting the ESBuild record 30 seconds in the future makes the cleanup cutoff unsafe. On a forced rebuild or branch-change rebuild within that buffer, `preBuildRecord.Timestamp` can be later than the chunk files just emitted, and `CleanupOldChunks` can delete fresh chunks because their mtimes are before the future cutoff. Please store the real completion time and keep any recent-build suppression buffer separate, or clamp the cleanup cutoff so it can never be after the current build start/time.

build-tools/utilities/ProcessRunner.cs
```diff
@@ -22,7 +25,7 @@ public static Task RunNpmCommand(string workingDirectory, string command,
     {
         return RunCommand(workingDirectory, "pwsh", "-Command",
```
TimPurdum, line 26: Copilot/GPT-5.5: The new shared npm runner hard-requires `pwsh` for every npm command. The Docker images added in this PR install Node but not PowerShell, and many Linux/macOS developer machines will not have `pwsh` available, so the ESBuild step fails as soon as it reaches `npm install`/`npm run`. Since `npm` is already an executable, please invoke `npm` directly with argument tokens, or install/document `pwsh` in every environment that uses these build tools.

src/dymaptic.GeoBlazor.Core.ESBuild/dymaptic.GeoBlazor.Core.ESBuild.csproj
```diff
@@ -3,8 +3,8 @@
         <TargetFrameworks>net8.0;net10.0;netstandard2.0</TargetFrameworks>
     </PropertyGroup>
     <!-- ESBuild-Triggering "Project" to enforce that this only fires once per build, no matter what project we build from -->
-    <Target BeforeTargets="DispatchToInnerBuilds" Name="ESBuild">
-        <Exec Command="dotnet $(GBBuildToolsPath)/ESBuild.dll -c $(Configuration) --dialog" ContinueOnError="false" Condition="$(ShowScriptDialogs) == 'true'"/>
-        <Exec Command="dotnet $(GBBuildToolsPath)/ESBuild.dll -c $(Configuration)" ContinueOnError="false" Condition="$(ShowScriptDialogs) != 'true'"/>
+    <Target BeforeTargets="DispatchToInnerBuilds" Name="ESBuild" Condition="'$(SkipESBuild)' != 'true'">
+        <Exec Command="dotnet $(GBBuildToolsPath)/ESBuild.dll -c $(Configuration)$(CoreESBuildLint) --dialog" ContinueOnError="false" Condition="$(ShowScriptDialogs) == 'true'" StandardOutputImportance="high" StandardErrorImportance="high"/>
```
TimPurdum: Copilot/GPT-5.5: `$(GBBuildToolsPath)` and `$(Configuration)` are unquoted in this `Exec` command. That breaks checkouts under paths with spaces and also breaks the new `SourceGen Highlighting` configuration because `-c SourceGen Highlighting` is passed as two arguments. Please quote both values, e.g. `dotnet &quot;$(GBBuildToolsPath)/ESBuild.dll&quot; -c &quot;$(Configuration)&quot; ...`, and apply the same fix to the non-dialog command below.

Directory.Build.props
```diff
@@ -66,25 +73,32 @@
...
+        <Exec Command="dotnet &quot;$(GBBuildToolsPath)/AcquireBuildLock.dll&quot; &quot;$(ESBuildLockFilePath)&quot; 5"
+              IgnoreExitCode="true">
+            <Output TaskParameter="ExitCode"
+                    PropertyName="_AcquireLockExitCode" />
+        </Exec>
+        <!-- Lock acquired: run ESBuild, then remove the .running marker -->
+        <Message Importance="high"
+                 Text="Triggering Core ESBuild from $(MSBuildProjectName)"
+                 Condition="'$(_AcquireLockExitCode)' == '0'" />
+        <Exec Command="dotnet restore $(MSBuildThisFileDirectory)src/dymaptic.GeoBlazor.Core.ESBuild/dymaptic.GeoBlazor.Core.ESBuild.csproj"
```
TimPurdum: Copilot/GPT-5.5: This `dotnet restore` command leaves the project path unquoted, so restore fails when the repository lives under a path containing spaces. Please quote the path (`dotnet restore &quot;$(MSBuildThisFileDirectory)src/...csproj&quot;`) or use an MSBuild task with `Targets="Restore"` to avoid shell tokenization.

TimPurdum:
Copilot/GPT-5.5 review follow-up: a few additional findings were valid but GitHub rejected their inline anchors, so I am posting them here instead.

1. `src/dymaptic.GeoBlazor.Core.SourceGenerator/dymaptic.GeoBlazor.Core.SourceGenerator.csproj`: `TargetPathWithTargetPlatformMoniker` reconstructs the shared source-generator DLL path as `bin\$(Configuration)\netstandard2.0`. That breaks with custom output layouts such as `--artifacts-path`, `BaseOutputPath`, or `OutputPath`, and can point at a stale/missing assembly in clean builds. Please flow the referenced project resolved target path through MSBuild items instead of hardcoding `bin`.

2. `test/dymaptic.GeoBlazor.Core.SourceGenerator.Tests/dymaptic.GeoBlazor.Core.SourceGenerator.Tests.csproj`: the new references to `..\..\src\dymaptic.GeoBlazor.Core\bin\Debug|Release\net8.0\dymaptic.GeoBlazor.Core.dll` make clean `dotnet test` depend on Core already being built in exactly that path/configuration and ignore custom output/artifacts paths. Please use a `ProjectReference` or resolve the target path via MSBuild.

3. `Dockerfile.unit`: `COPY ./test/Directory.Build.* ./test/Directory.Build.*` uses a wildcard in the destination. Docker expands wildcards only in sources, so this copies `test/Directory.Build.props` to a literal file named `Directory.Build.*`; MSBuild will not import it. The destination should be `./test/`.

4. `Dockerfile` and `dev-pr-build.yml`: both call `./build-tools/linux-x64/GeoBlazorBuild.dll`, but this PR only tracks the `build-tools/win-x64` build-tool binaries. The `linux-x64` directory is not present in the tracked tree with the required DLL/runtimeconfig files, so clean Linux/Docker builds will fail unless the Linux build-tool outputs are generated before those calls or committed.

5. `Dockerfile` / `Dockerfile.unit`: both still contain `COPY ./*.ps1 ./`, but the PR deletes the root `GeoBlazorBuild.ps1` and there are no root `.ps1` files left. Docker fails on unmatched source globs, so these lines should be removed or changed to copy existing files only.

build-tools/utilities/ProcessRunner.cs
TimPurdum: This shouldn't be an issue since we're using dotnet images (which always include pwsh, as I understand it).

TimPurdum (commented):
## 🤖 Follow-up Review by Claude (Opus 4.8)

_Re-review of the fixes pushed after [my first review](https://github.com/dymaptic/GeoBlazor/pull/521#pullrequestreview-4450287860), comparing `96ef1b37` → `6f3a11b4`. Advisory, non-blocking._

### ✅ Verified fixed
| # | Finding | Fix | Verdict |
|---|---------|-----|---------|
| 1 | Light-theme regression (`setComponentTheme`) | `theme === 'dark'` guard restored on both branches | ✅ Correct |
| 2 | `ProcessHelper.Execute` unbounded retry + dead kill | `MaxRetryAttempts = 5` cap added; kill now matches via `TryGetProcessCommandLine` | ✅ Bounded (see residual below) |
| 3 | `ESBuildGenerator` fire-and-forget context escape | Now runs synchronously via `.GetAwaiter().GetResult()` with explanatory comment | ✅ Correct |
| 6 | `MapColor.GetHashCode` reference-vs-value | Now folds `RgbaValues` contents + `HexOrNameValue` into a `HashCode` | ✅ Consistent with `Equals` |

Nice — the four highest-value items are addressed cleanly.

### ⚠️ Residuals / still open
- **#2 residual (Windows):** `TryGetProcessCommandLine` only implements Linux (`/proc`) and macOS (`ps`); on **Windows it returns `""`**, so `"".Contains(programName)` is always false and the lock-holder is never killed on the primary dev platform. Not dangerous now (the retry is bounded and the 500 ms delay may let the lock clear on its own), but the remediation the fix intends doesn't actually run on Windows. Inline note added.
- **#7 still open:** the non-generic-`Task` return-type parse in `SerializationGenerator` is unchanged (the only edit to that file was a `new string(...)` → `new(...)` cleanup). Inline note re-anchored.
- **#4 not addressed:** `ProcessKiller.KillByProcessOrFileName` still uses a substring match + kill-all (can terminate unrelated `dotnet`/`node` processes). Untouched.
- **#5 not addressed:** `ProcessRunner.RunCommand` still orphans the child process when the `CancellationToken` fires (kill is unreachable past `WaitForExitAsync`). Untouched.

### Note on the theme fix
The initial-setup bug is fixed. One edge remains: the guard now only *adds* `calcite-mode-dark` for dark and never *removes* it, so a runtime **dark → light** toggle leaves already-dark widgets stuck dark. If runtime theme switching is supported, `classList.toggle('calcite-mode-dark', theme === 'dark')` would cover both directions. Inline note added.

#4, #5, and #7 were Low/Medium and are fine to defer to a follow-up issue if you'd rather not expand this PR further.

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/ProcessHelper.cs
```diff
@@ -438,6 +455,49 @@ private static bool HasProcessExited(Process process)
...
+
+    /// <summary>
+    ///     Attempts to read the full command-line string of an external process in a cross-platform manner.
+    ///     Returns an empty string if the command line cannot be determined.
+    /// </summary>
+    private static string TryGetProcessCommandLine(int pid)
+    {
+        try
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
TimPurdum, line 468: **🤖 Claude follow-up — residual on the retry fix (Windows).**

The bounded-retry + command-line-match rewrite is a good improvement, but `TryGetProcessCommandLine` only branches for `Linux` (here) and `OSX` below — there's no `Windows` branch, so on Windows it falls through to `return string.Empty`. That makes the lock-holder filter (`TryGetProcessCommandLine(p.Id).Contains(programName)` at line ~223) always false on Windows, so the kill never fires on the primary dev platform.

The retry is now safely bounded, so this isn't a hang risk — but if you want the kill to actually clear the lock on Windows, you'd need a Windows command-line lookup (e.g. WMI `Win32_Process.CommandLine` via `System.Management`, or CIM). Otherwise consider dropping the kill attempt and relying purely on the delay+retry, to avoid implying remediation that doesn't run.

src/dymaptic.GeoBlazor.Core/Scripts/arcGisJsInterop.ts
```diff
@@ -178,19 +181,39 @@ export function setTheme(theme: string | null, viewId): string | null {
...
+            if ((component instanceof MapView || component instanceof SceneView)) {
+                component!.container!.style.colorScheme = theme as string;
+                if (theme === 'dark' && !component!.ui.container!.classList.contains('calcite-mode-dark')) {
+                    // if the view was already rendered, this class is missed and needs adding
+                    component!.ui.container!.classList.add('calcite-mode-dark');
+                }
+            }
+
+            if (component instanceof HTMLElement) {
+                if (theme === 'dark' && !component.classList.contains('calcite-mode-dark')) {
```
TimPurdum, line 212: **🤖 Claude follow-up — theme fix verified, one edge remains.**

The `theme === 'dark'` guard correctly fixes the initial light-theme bug. The remaining edge: this only *adds* `calcite-mode-dark` and never removes it, so a runtime **dark → light** switch leaves a previously-dark widget stuck dark. If `setTheme` can be called to switch to light at runtime, `component.classList.toggle('calcite-mode-dark', theme === 'dark')` (and the same on the MapView/SceneView `ui.container` above) handles both directions in one line. Non-blocking if runtime light-switching isn't a supported scenario.

src/dymaptic.GeoBlazor.Core.SourceGenerator.Shared/SerializationGenerator.cs
```diff
@@ -515,7 +514,7 @@ private static string GenerateSerializableParameterRecord(string value, int inde
             ? "null"
```
TimPurdum, line 514: **🤖 Claude follow-up — #7 still open** _(anchored here; the actual line is ~544, just outside the diff hunk)._

The non-generic-`Task` return parse is unchanged from the first review. At line ~544, for a bare `Task`/`ValueTask` (no `<…>`), `IndexOf('<')` returns `-1`, so `Substring(bracketIndex + 1, returnType.Length - bracketIndex - 2)` becomes `Substring(0, Length - 1)` and truncates the type name (`"Task"` → `"Tas"`). Guard with `if (bracketIndex < 0)` and treat it as void/no-return. Low priority (masked today because serialized methods return `Task<T>`), just flagging it wasn't picked up in the fix pass.

Directory.Build.props
```diff
@@ -66,25 +73,32 @@
...
+            <Output TaskParameter="ExitCode"
+                    PropertyName="_AcquireLockExitCode" />
+        </Exec>
+        <!-- Lock acquired: run ESBuild, then remove the .running marker -->
+        <Message Importance="high"
+                 Text="Triggering Core ESBuild from $(MSBuildProjectName)"
+                 Condition="'$(_AcquireLockExitCode)' == '0'" />
+        <Exec Command="dotnet restore &quot;$(MSBuildThisFileDirectory)src/dymaptic.GeoBlazor.Core.ESBuild/dymaptic.GeoBlazor.Core.ESBuild.csproj&quot;"
+              Condition="'$(_AcquireLockExitCode)' == '0'" />
+        <MSBuild Projects="&quot;$(MSBuildThisFileDirectory)src/dymaptic.GeoBlazor.Core.ESBuild/dymaptic.GeoBlazor.Core.ESBuild.csproj&quot;"
```
TimPurdum: Copilot/GPT-5.5 follow-up: this attempted quoting fix breaks the `MSBuild` task. `Projects` is an MSBuild item value, not a shell command, so the `&quot;...&quot;` characters become part of the project path; a local MSBuild repro fails with MSB3202 looking for a file whose name includes literal quotes. Please remove the quotes here and keep quoting only on the `Exec` command above.

TimPurdum:
Copilot/GPT-5.5 follow-up review:

1. `.github/workflows/main-release-build.yml`: the `build` job now runs on `[ self-hosted, Windows, X64 ]`, but the `Install xmllint` step still runs `sudo apt update` / `sudo apt install`, which will fail on Windows. The following `Copy Build Version` step also has no `shell: bash` and uses bash syntax (`CORE_VERSION=$(...)`, `>> $GITHUB_ENV`) under the Windows default shell. Please either move this job back to Ubuntu or rewrite those steps as PowerShell/Windows-compatible commands.

2. `src/dymaptic.GeoBlazor.Core.ESBuild/dymaptic.GeoBlazor.Core.ESBuild.csproj`: the path is now quoted, but `$(Configuration)` is still not. The repo defines `SourceGen Highlighting`, so ESBuild still receives `-c SourceGen Highlighting` as two arguments. Please use `-c &quot;$(Configuration)&quot;` on both Exec commands.

3. `build-tools/build-scripts/ESBuild.cs`: the future-dated record timestamp is still present (`DateTimeOffset.UtcNow.AddSeconds(30)`). That keeps the cleanup cutoff bug from the earlier review: a rebuild inside the buffer can delete freshly emitted chunks whose mtimes are earlier than the future cutoff.

4. `build-tools/utilities/ProcessRunner.cs`: `RunNpmCommand` still shells through `pwsh`. The Linux .NET SDK images used by the Dockerfiles do not include PowerShell by default, so Docker ESBuild paths still fail unless PowerShell is installed or npm is invoked directly.

5. `test/dymaptic.GeoBlazor.Core.SourceGenerator.Tests.csproj`: the hardcoded references to `src/dymaptic.GeoBlazor.Core/bin/Debug|Release/net8.0/dymaptic.GeoBlazor.Core.dll` are still present, so clean tests/custom output paths can still miss the DLL. Please replace with a project reference or MSBuild-resolved target path.

PR #530, opened 2026-06-17
Restore SnappingOptions runtime setters (4.5.x regression)

magmoe:
## What
Restores the runtime setter methods on `SnappingOptions` that were accidentally dropped in the 4.5.x line, plus regression guards.

## Why
Commit `c3ab36411` truncated `SnappingOptions.gb.cs` right after the property getters, deleting the entire Property Setters and collection-helper regions. That removed `SetEnabled`, `SetGridEnabled`, `SetDistance`, `SetFeatureEnabled`, `SetSelfEnabled`, `SetAttributeRulesEnabled`, and `SetFeatureSources`, so consumers can no longer toggle snapping / grid snapping at runtime (reported against 4.5.1). It's isolated to this one component and is still broken on `develop` and in 4.5.1–4.5.3.

## Changes
- Restore the complete `SnappingOptions.gb.cs` from `f02afefeb` (the commit immediately before the truncation) — re-adds exactly the 7 setters + the Add/Remove `FeatureSources` helpers and nothing else.
- `SnappingOptionsApiTests` — public-API contract test so a future code-generation split can't silently drop these members again.
- `GeometryExtentSerializationTests` + additions to `GeometryEngineTests.TestUnion` — guards for the separately-reported `GeometryEngine.Union` extent issue. That one is **not** reproducible from code (the union/extent path is identical to the working 4.4.4), so these lock in the behavior rather than fix a found bug.

## Notes
- Targets `develop` because the regression is present there too; also serves as the cut point for a 4.5.4 hotfix.
- 11/11 unit tests pass; the browser test project compiles.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TimPurdum:
@magmoe did you or Claude do a search for any similarly truncated files? It's possible I manually truncated this file to fix a merge conflict, but I want to make sure it's not a larger issue.

magmoe:
> @magmoe did you or Claude do a search for any similarly truncated files? It's possible I manually truncated this file to fix a merge conflict, but I want to make sure it's not a larger issue.

Yes, it's isolated to the one file. Claude checked to make sure

TimPurdum:
Added the source fix behind the Union/Extent regression guards on this branch (commit c5f7b16) — these were previously assertion-only, so `TestUnion`'s extent check would have failed without a fix.

**Root cause:** results from the geometry *operator* API (e.g. `unionOperator.executeMany`) don't always expose a cached `.extent` the way the legacy `geometryEngine.union` did. `buildDotNetPolygon`/etc. only copy `extent` when present, so the .NET `Geometry.Extent` came back `null` (breaking `map.GoTo(extent)`), and `GetExtent()` couldn't recover it because the result isn't a registered JS component. The C#/proto serialization itself was fine — the extent simply never made it into the payload.

**Fix (centralized, so every operator benefits — not just Union):**
- `Scripts/geometry.ts` — `buildDotNetGeometry` now populates a missing `extent`, computing a bounding box from the geometry's `rings`/`paths`/`points` when the geometry doesn't provide one.
- `Geometry.GetExtent()` — falls back to a client-side `CalculateExtent()` (overridden by `Polygon`/`Polyline` via a shared `CalculateExtentFromPaths`) when neither the cached value nor the JS component supplies one. Also no longer overwrites a good cached extent with a `null` from a stray JS call.

**Tests:**
- Added extent assertions to `TestUnionWithParamsArray` (mirrors your `TestUnion`).
- New `GeometryExtentTests` unit test covering the C# calculation fallback (`GetExtent()` with no JS module) — complements your `GeometryExtentSerializationTests`.

**Verification:** 21/21 unit tests pass on the branch; C#/TS compile clean (ESLint passed). I could **not** run the browser automation suite (`TestUnion`/`TestUnionWithParamsArray`) in my environment — the WASM runner wouldn't start — so those two e2e tests should be run in CI before merge.

Your `SnappingOptions` setters, contract tests, and version bump are untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

TimPurdum (commented):
Copilot: I found one edge case that should be addressed before merging.

In `src/dymaptic.GeoBlazor.Core/Components/Geometries/Geometry.cs` around `CalculateExtentFromPaths`, and the matching fallback in `src/dymaptic.GeoBlazor.Core/Scripts/geometry.ts`, the newly calculated extent only carries `xmin/ymin/xmax/ymax`. For paths/rings/points with Z or M coordinates, the fallback drops `zmin/zmax` and `mmin/mmax`, even though GeoBlazor's `Extent` model and ArcGIS `Extent` both support those values. That makes the fallback less equivalent to `geometry.extent` for Z/M-enabled geometry results, and can return an incomplete extent for 3D/M geometries when the ArcGIS cached extent is missing.

Suggested fix: while scanning vertices, also track coordinate index 2 when `hasZ` is true and the M index (`3` when `hasZ`, otherwise `2`) when `hasM` is true, then populate `Zmin/Zmax/Mmin/Mmax` in C# and `zmin/zmax/mmin/mmax` in TypeScript. Adding a regression test with 3D or M-valued polygon/polyline coordinates would lock this down.

TimPurdum:
Addressed the Z/M edge case in `fe23867e5`.

`CalculateExtentFromPaths` (Geometry.cs) and `calculateGeometryExtent` (geometry.ts) now also track `zmin/zmax` and `mmin/mmax` when the geometry has Z/M — z at coordinate index 2 when `hasZ`, m at index 3 when `hasZ` else index 2 — and populate the resulting extent's Z/M bounds (and `hasZ`/`hasM` flags). When the geometry exposes its own `.extent`, that's still used directly and is unaffected.

Added regression coverage in `GeometryExtentTests`: a Z-valued polygon (asserts `Zmin`/`Zmax`) and an M-valued polyline (asserts `Mmin`/`Mmax`). All 4 `GeometryExtentTests` pass and the TS recompiles clean (ESLint OK).

The remaining Copilot inline note on `SetFeatureSources` (calling `UpdateGeoBlazorReferences(CoreJsModule!...)` before the null-check) is the standard generated pattern shared by every collection setter, not specific to this PR — worth fixing in the code generator rather than hand-editing the `.gb.cs` here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

TimPurdum:
Reworked the extent fix (`0f42a874e`) per feedback: missing extents are now computed by **ArcGIS**, not by hand-rolled C#/TS bounding-box logic.

- `geometry.ts` `buildDotNetGeometry` — when an operator result (e.g. `unionOperator`) doesn't expose a cached `.extent`, it rebuilds a typed ArcGIS geometry via `buildJsGeometry` so the SDK recomputes the extent (Z/M bounds included automatically). It uses the geometry's own `.extent` when present.
- Reverted the C# `CalculateExtent` / `CalculateExtentFromPaths` and the `Polygon`/`Polyline` overrides; `Geometry.GetExtent()` is back to its original form (the extent now arrives populated from the JS side).
- Removed the C# `GeometryExtentTests` (those exercised the now-deleted C# calculation). Coverage remains via `GeometryExtentSerializationTests` and the browser `TestUnion`/`TestUnionWithParamsArray` extent assertions.

Verified: Core compiles, 19/19 unit tests pass, TS recompiles clean (ESLint OK). The browser e2e still needs a CI run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #534, opened 2026-06-23
Release v4.6.1

TimPurdum:
# Release v4.6.1

## Highlights

- Patch release for GeoBlazor Core with no public API changes or breaking changes.
- Fixes ESBuild incremental checks so CI branch metadata does not force unnecessary rebuilds.
- Fixes runtime layer updates for components backed by raw ArcGIS JavaScript objects, including `SnappingOptions` used by Pro widgets.

## Bug Fixes

- Fixed Core build script handling for ESBuild incremental checks so branch-name noise in CI does not invalidate JavaScript build state.
- Fixed `MapComponent.SetLayer` so it no longer calls a missing `setLayer` method directly on raw ArcGIS objects. This resolves runtime failures seen when updating `SnappingOptions` after render, including calls such as `SetEnabled(true)` in a Sketch workflow.

## Internal / Build / Tooling

- Refreshed bundled GeoBlazor build-tool binaries for Windows, Linux, and macOS.
- Synced `develop` from `main` before release so this PR includes the published v4.6.0 baseline.


TimPurdum:
Release note verification: compared v4.6.0..develop after syncing main into develop. There are no changes under src/, so no public API surface changed in this patch; the GeoBlazorUpdateTool API diff was not rerun because this release contains only build-tool updates plus the 4.6.1 version bump.

